PR #25, opened 2020-01-09
Rewrite functions for FHIR Renderer for all classes using smart-on-fhir client

zxenia:
- Used [smart-on-fhir Python client](https://github.com/smart-on-fhir/client-py) to rewrite converter functions for all classes.
- The client provides validation against FHIR 4.0.
- Important: update of fhir client might be needed in the future when the official version 4 is released, see [issue](https://github.com/smart-on-fhir/client-py/issues/70)
- Added tests


davidlougheed:
can you also add the new dependencies to setup.py?

PR #51, opened 2020-01-29
Fix auto now / auto now add flip

davidlougheed:


davidlougheed:
See https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.DateField

PR #64, opened 2020-02-17
Small change to data model

zxenia:
Added field 'version' to dataset with default value as "version_datetime.now()"


chord_metadata_service/chord/models.py
```diff
@@ -116,6 +117,8 @@ def n_of_tables(self):
                               help_text="The grant(s) which funded and supported the work reported by the dataset.")
     keywords = ArrayField(JSONField(null=True, blank=True), blank=True, null=True,
                           help_text="Tags associated with the dataset, which will help in its discovery.")
+    version = models.CharField(max_length=200, blank=True, default=f"version_{datetime.datetime.now()}",
```
afrosimon: I believe this will become a problem, the default value being interpreted once, when the model is loaded in the application. For more info : https://stackoverflow.com/questions/2771676/django-datetime-issues-default-datetime-now

I think a potential fix could be to create a lambda which will produce the string "version_..."

zxenia: The idea of default value I had in mind here:  if the version is not provided, then add version name with a datestamp when the dataset was created - creating a string 'version_{of this date}', and this value should stay persistent or the data provider can provide their own version.

afrosimon: Yes that's fine except if the version is not provided, the default will stay the same until the uwsgi or gunicorn server is reloaded, as in, default will be "version_{value of datetime.now() at the moment this module is imported}".

So default should receive a method for this value to reflect the actual datetime.now() every time a default value is needed otherwise currently it's merely a string, which will stay the same : https://docs.djangoproject.com/en/3.0/ref/models/fields/#default

chord_metadata_service/chord/migrations/0008_dataset_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('chord', '0007_auto_20200129_1537'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='dataset',
+            name='version',
+            field=models.CharField(blank=True, default='version_2020-02-17 16:47:59.425036', help_text='A release point for the dataset when applicable.', max_length=200),
```
afrosimon: The default value is fixed in the model but I believe you'd need to recreate / update this migration accordingly.

zxenia: done!

PR #75, opened 2020-02-24
Fix for configuration of requirements and version for docs build

zxenia:


docs/conf.py
```diff
@@ -25,9 +25,9 @@
 author = 'Ksenia Zaytseva, David Lougheed, Simon Chenard'
 
 # The full version, including alpha/beta/rc tags
-release = '0.3.0'
-
+release = '0.4.0'
 
+version = '0.4.0'
```
davidlougheed: If this is the project version, we should make a release checklist or find a way to sync this with setup.py

zxenia: I can remove it for now till we have a decision

PR #97, opened 2020-03-20
Experiments service

romgrk:


chord_metadata_service/experiments/models.py
```diff
@@ -0,0 +1,62 @@
...
+
+    id = models.CharField(primary_key=True, max_length=200, help_text='An arbitrary identifier for the individual.')
+
+    reference_registry_id = models.CharField(max_length=30, null=True, blank=True,
+            help_text='The IHEC EpiRR ID for this dataset, only for IHEC Reference Epigenome datasets. Otherwise leave empty.')
+    qc_flags = ArrayField(models.CharField(max_length=30, null=True, blank=True,
+            help_text='Any quanlity control observations can be noted here. This field can be omitted if empty'))
+    experiment_type = models.CharField(max_length=30, null=True, blank=True,
+            help_text="(Controlled Vocabulary) The assay target (e.g. ‘DNA Methylation’, ‘mRNA-Seq’, ‘smRNA-Seq’, 'Histone H3K4me1').")
+    experiment_ontology_curie = ArrayField(models.CharField(max_length=30, null=True, blank=True,
```
zxenia: I think we should store ontology terms in the same format we store them in other apps as:
```
{
    "id": "some_id",
    "label": "some_label"
 }
```
Then the items in ArrayField should be of JSONField type, and no regex needed.
We won't store just curies, so the field also can be renamed to `experiment_ontology`

chord_metadata_service/experiments/models.py
```diff
@@ -0,0 +1,62 @@
...
+            validators=[RegexValidator(regex="^[A-Za-z]*:[A-Za-z0-9]*")],
+            help_text="(Ontology: OBI) links to experiment ontology information."))
+    molecule_ontology_curie = ArrayField(models.CharField(max_length=30, null=True, blank=True,
```
zxenia: same as above

chord_metadata_service/experiments/models.py
romgrk: Can I make it simply a JSONField that would contain an array of items? Validation will be more performant that way imho.

zxenia: You can try, but I think Postgres will validate the JSONField, so you will need some property to store a list in.

romgrk:
@zxenia for `qc_flags` you proposed 500 max-length, but I had missed the array-type. So what do you think about an array of max-length-100 strings? Or do you want to keep an array of max-length-500 strings?

zxenia:
@romgrk array of max-length-100 - sounds good!

chord_metadata_service/experiments/models.py
davidlougheed: what do you mean by needing a property? a straight array is valid JSON too, so it should be OK

romgrk:
And for the `experiment_type` should we implement validation to only accept the experiment types permitted in the IHEC experiment jsonschema?

zxenia:
good question! it states in ihec schema: `The assay target (e.g. ‘DNA Methylation’, ‘mRNA-Seq’, ‘smRNA-Seq’, 'Histone H3K4me1'` but does this mean that only these values are allowed? I think it is just an example of values and the actual list is actually longer.

romgrk:
Weird bug on my side: JSONField's validators don't seem to be called. Is it just me?

chord_metadata_service/experiments/models.py
zxenia: oh yeah it's true, it should be ok

romgrk:
The permitted experiment_types are these ones:
```python
# This one is the top-level one
#  "experiment_type" : {"type": "string", "description": "(Controlled Vocabulary) The assay target (e.g. ‘DNA Methylation’, ‘mRNA-Seq’, ‘smRNA-Seq’, 'Histone H3K4me1')."},

# And these are the ones on each experiment subtype
#  "experiment_type" : {"type": "string", "enum": ["Chromatin Accessibility"]},
#  "experiment_type" : {"type": "string", "enum": ["DNA Methylation"]},
#  "experiment_type" : {"type": "string", "enum": ["DNA Methylation"]},
#  "experiment_type" : {"type": "string", "enum": ["DNA Methylation"]},
#  "experiment_type" : {"type": "string", "pattern": "^(ChIP-Seq Input)|(Histone H\\w+([\\./]\\w+)?)+$|(Transcription Factor)"},
#  "experiment_type" : {"type": "string", "enum": ["RNA-Seq", "mRNA-Seq", "smRNA-Seq", "total-RNA-Seq"]},
#  "experiment_type" : {"type": "string", "enum": ["WGS"]},
#  "experiment_type" : {"type": "string", "enum": ["Other"]},
```

zxenia:
I think we should use this list then for validation.

romgrk:
I've updated the PR to match the coding style used in other models, and I've added validation. @zxenia could you take a look and approve the changes?

Remaining questions:
 - Should `experiment_type` be required?
 - Should `experiment_type` be validated?
 - Using `Model.objects.create(...)` seems to create an object in the DB without calling validators. `m = Model(...); m.full_clean(); m.save()` seems to do the validation+save. Is this what we expect?

zxenia:
@romgrk 

- I think `experiment_type` should be required, it is  required in ihec spec
- seems so -  `full_clean()` is not called automatically so you have to call it manually  [docs](https://docs.djangoproject.com/en/2.2/ref/models/instances/#django.db.models.Model.full_clean)

chord_metadata_service/experiments/models.py
```diff
@@ -0,0 +1,56 @@
...
+        ('genomic DNA', 'genomic DNA'),
+        ('protein', 'protein'),
+        ('other', 'other'),
+    )
+
+    id = CharField(primary_key=True, max_length=200, help_text=rec_help(d.EXPERIMENT, 'id'))
+
+    reference_registry_id = CharField(max_length=30, null=True, blank=True, help_text=rec_help(d.EXPERIMENT, 'reference_registry_id'))
+    qc_flags = ArrayField(CharField(max_length=100, help_text=rec_help(d.EXPERIMENT, 'qc_flags')), null=True, default=list)
+    experiment_type = CharField(max_length=30, null=True, blank=True, help_text=rec_help(d.EXPERIMENT, 'experiment_type'))
```
zxenia: let's make it required field

chord_metadata_service/experiments/models.py
```diff
@@ -0,0 +1,56 @@
...
+    experiment_ontology = JSONField(null=True, blank=True, validators=[ontologyListValidator], help_text=rec_help(d.EXPERIMENT, 'experiment_ontology'))
```
zxenia: let's make it required field

romgrk:
Ok, `experiment_type` has been made required.

Issue #98, opened 2020-03-23
Ontology class vs. CodeableConcept


zxenia:
Currently, for all ontology terms, we use the ONTOLOGY_CLASS schema that follows Phenopackets schema.
E.g.
```
"code": {
        "id": "NCIT:C28743",
        "label": "Punch Biopsy"
    }
```

mCode is using FHIR datatypes for the data elements. The concepts are encoded with FHIR CodeableConcept datatype.

E.g.

```
"code": {
        "coding": [
            {
              "code": "50610-5",
               "display": "Alpha-1-Fetoprotein",
               "system": "https://loinc.org/"
             }
         ]
}
```

**Issue:**
 The ontology term representation should be unified across metadata service. Otherwise, I think it can be confusing to have two schemas for Concepts.

Phenopackets ONTOLOGY_CLASS doesn't have a` 'system'` and free `'text'` attribute as FHIR Concept.
Our current ontology schema doesn't allow for `additionalProperties`.

**Solution:**
Set `"additionalProperties": True` to allow for 'system' and 'text' attribute, keep required list as before - ['id', 'label']

**Pros:** one schema for all ontology terms across metadata service

**Cons:** anything can be added besides id and label

zxenia:
implemented, using ONTOLOGY_CLASS schema across all the apps

PR #103, opened 2020-03-27
Mcode app

zxenia:
Added a new app to store more detailed cancer-related metadata based on [mCode](https://mcodeinitiative.org/) data elements.

**Main features:**

- mCode uses FHIR datatypes for each data element. I decided not to use FHIR native CodeableConcept for representing any concept term because it is very complex and there is already Ontology_class schema for handling concepts that are taken from Phenopackets (#98). Except of concept, other FHIR datatypes were used in schemas.

- Extended Individual class with more fields from mCode (ecog_performance_status, karnofsky, comorbid_condition).

- Set validators in models only in mCode

- Updated JsonSchemaValidator in order to allow setting format_checker for the value

- Added mcodepacket json example in `/example` dir

**Remaining questions:**
1. How to align Cancer Condition and Disease in Phenopackets? The classes described quite differently: Disease has a property 'tnm_finding' (ontology term). Cancer condition in mcode has a relation to TNMStaging that described as much detailed.

2. How to deal with Variants? Different description in both apps.


chord_metadata_service/experiments/migrations/0002_auto_20200326_1339.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 2.2.10 on 2020-03-26 17:39
+
+import chord_metadata_service.restapi.validators
```
zxenia: this migration was generated by django because the JsonSchemaValidator was updated but there are no actual changes to the models. Should I discard it?

afrosimon: If you downgrade, get rid of it and then generate new migrations, is it there (based on your description, it shouldn't)?

zxenia: I am just thinking that it can collide with migrations on Romain's branch because he is working with experiments app: it means there will be two migration files that have a dependency to an initial migration, and both files will have 0002 sequence number. So when someone checks out the code and migrates how django knows what migrations apply first? 
that's why maybe just delete this useless migration :)

afrosimon: Ah right, tricky situation for sure. In this case sounds like you can simply delete it. But there's also the option of squashing multiple migrations together, if it comes to that.

chord_metadata_service/mcode/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
davidlougheed: remove this file, since you have a tests module in folder form

setup.py
```diff
@@ -33,6 +33,7 @@
         "rdflib-jsonld==0.4.0",
         "requests>=2.23,<3.0",
         "uritemplate>=3.0,<4.0",
+        "rfc3987==1.3.8",
```
davidlougheed: ideally we can keep these in alphabetical order for clarity. not a huge deal though

chord_metadata_service/mcode/api_views.py
```diff
@@ -0,0 +1,57 @@
...
+from chord_metadata_service.restapi.pagination import LargeResultsSetPagination
+
+
+class McodeModelViewSet(viewsets.ModelViewSet):
+    pagination_class = LargeResultsSetPagination
+    renderer_classes = (*api_settings.DEFAULT_RENDERER_CLASSES, PhenopacketsRenderer)
+
+
+class GeneticVariantTestedViewSet(McodeModelViewSet):
+    queryset = GeneticVariantTested.objects.all().order_by("id")
```
afrosimon: Bit of a nitpick but I'd say the default ordering would be better located in the model itself (https://docs.djangoproject.com/en/3.0/ref/models/options/#ordering), possibly in a mixin if we keep using the same field to order by.

chord_metadata_service/mcode/models.py
```diff
@@ -0,0 +1,276 @@
...
+)
+
+
+############################ Field validators #############################
+
+ontology_class_validator = JsonSchemaValidator(ONTOLOGY_CLASS)
+quantity_validator = JsonSchemaValidator(schema=QUANTITY, format_checker=['uri'])
+tumor_marker_test_validator = JsonSchemaValidator(schema=TUMOR_MARKER_TEST)
+complex_ontology_validator = JsonSchemaValidator(schema=COMPLEX_ONTOLOGY, format_checker=['uri'])
+time_or_period_validator = JsonSchemaValidator(schema=TIME_OR_PERIOD, format_checker=['date-time'])
```
romgrk: I've already defined some of these in `validators.py` in the experiments PR that you just merged.

chord_metadata_service/mcode/models.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class GenomicsReport(models.Model, IndexableMixin):
+    """
+    Genetic Analysis Summary
+    """
+
+    id = models.CharField(primary_key=True, max_length=200, help_text=rec_help(d.GENOMICS_REPORT, "id"))
+    test_name = JSONField(validators=[ontology_class_validator], help_text=rec_help(d.GENOMICS_REPORT, "test_name"))
+    performing_ogranization_name = models.CharField(max_length=200, blank=True,
```
romgrk: typo `ogranization` => `organization`

chord_metadata_service/mcode/models.py
```diff
@@ -0,0 +1,276 @@
...
+    """
+    CANCER_CONDITION_TYPE = (
+        ('primary', 'primary'),
+        ('secondary', 'secondary')
+    )
+    id = models.CharField(primary_key=True, max_length=200, help_text=rec_help(d.CANCER_CONDITION, "id"))
+    condition_type = models.CharField(choices=CANCER_CONDITION_TYPE, max_length=200,
+                                      help_text=rec_help(d.CANCER_CONDITION, "condition_type"))
+    # TODO add body_location_code validator array of json
+    body_location_code = ArrayField(JSONField(null=True, blank=True, validators=[ontology_class_validator]),
```
romgrk: This should be consistent with what I did in the experiments model, where we picked a JSONField containing an array rather than an array of JSONFields. We can review the decision but it should be consistent across models.

chord_metadata_service/mcode/models.py
```diff
@@ -0,0 +1,276 @@
...
+        ('radiation', 'radiation'),
+        ('surgical', 'surgical')
+    )
+    id = models.CharField(primary_key=True, max_length=200, help_text=rec_help(d.CANCER_RELATED_PROCEDURE, "id"))
+    procedure_type = models.CharField(choices=PROCEDURE_TYPES, max_length=200,
+                                      help_text=rec_help(d.CANCER_RELATED_PROCEDURE, "procedure_type"))
+    code = JSONField(validators=[ontology_class_validator], help_text=rec_help(d.CANCER_RELATED_PROCEDURE, "code"))
+    occurence_time_or_period = JSONField(validators=[time_or_period_validator],
+                                         help_text=rec_help(d.CANCER_RELATED_PROCEDURE, "occurence_time_or_period"))
+    target_body_site = ArrayField(JSONField(null=True, blank=True, validators=[ontology_class_validator]),
```
romgrk: Same as above

zxenia:
@romgrk they have different functionality, in our case, we only care about  `null=True` because this is what affects the database.
`blank=True` is used for validation in forms for data input, so when the field is set `blank=True` means the field is not required in forms.

chord_metadata_service/restapi/validators.py
```diff
@@ -31,3 +34,7 @@ def deconstruct(self):
 ontologyListValidator = JsonSchemaValidator(ONTOLOGY_CLASS_LIST)
 keyValueValidator = JsonSchemaValidator(KEY_VALUE_OBJECT)
 ageOrAgeRangeValidator = JsonSchemaValidator(AGE_OR_AGE_RANGE)
+quantity_validator = JsonSchemaValidator(schema=QUANTITY, format_checker=['uri'])
+tumor_marker_test_validator = JsonSchemaValidator(schema=TUMOR_MARKER_TEST)
+complex_ontology_validator = JsonSchemaValidator(schema=COMPLEX_ONTOLOGY, format_checker=['uri'])
+time_or_period_validator = JsonSchemaValidator(schema=TIME_OR_PERIOD, format_checker=['date-time'])
```
romgrk, line 40: Nitpick but variable names should be consistent, either all `camelCase` or all `snake_case`

zxenia: @romgrk yeah I know, it should be all `snake_case` for variables, can we do this in general code refactoring after mcode merged because it uses shared validators that are used in experiments, I don't want to introduce changes to experiments models on this branch

PR #122, opened 2020-05-15
Fixes to ingest

zxenia:
small fixes I had to do when ingesting signature data


chord_metadata_service/chord/views_ingest.py
```diff
@@ -153,13 +153,13 @@ def ingest(request):
 
 def _query_and_check_nulls(obj: dict, key: str, transform: Callable = lambda x: x):
     value = obj.get(key, None)
-    return {f"{key}__isnull": True} if value is None else {key: transform(value)}
+    return {f"{key}__isnull": True} if value is None or value == "" else {key: transform(value)}
```
davidlougheed, line 156: I don't understand why this is needed, since `""` isn't the same thing as null so `__isnull` shouldn't work here?

chord_metadata_service/chord/views_ingest.py
```diff
@@ -222,7 +222,7 @@ def ingest_phenopacket(phenopacket_data, table_id):
             term=disease["term"],
             disease_stage=disease.get("disease_stage", []),
             tnm_finding=disease.get("tnm_finding", []),
-            **_query_and_check_nulls(disease, "onset")
+            onset=disease.get("onset", None)
```
davidlougheed, line 225: `_query_and_check_nulls` is meant to do the `=None` check without causing issues

chord_metadata_service/chord/views_ingest.py
zxenia: Yes, the case was that somewhere it is an empty string in the data (e.g. `"date_of_birth": ""`) and it is not treated as None so it tries to transform.

chord_metadata_service/chord/views_ingest.py
zxenia: yeah, I see here, I don't understand why then it didn't save the "onset" value in db.

chord_metadata_service/chord/views_ingest.py
davidlougheed: maybe add a flag argument called "convert_empty_str" or something which defaults to false, and if true changes value to None if it's "" in
and then use that flag but only for the specific situation where you need it?

zxenia: is there a drawback of leaving it like this? usually, I wouldn't know if the data has empty strings and where (it can be all properties that accept string).

davidlougheed: the issue that can happen with this is that if a field can be either None or "" (unlikely but possible...) it will miss instances in the database where "" stored

PR #126, opened 2020-05-27
[WIP] FHIR to Phenopackets ingestion

zxenia:
- Current FHIR resources that are supported for ingestion are Patient, Observation, Condition, Specimen
- The ingestion works with [FHIR Bundle resource](https://www.hl7.org/fhir/bundle.html)
- The schemas for validation of an ingest body and FHIR Bundle are added
- For each new Patient, a Phenopacket is being created
- Examples of FHIR data added


chord_metadata_service/chord/data_types.py
```diff
@@ -9,6 +9,7 @@
 
 DATA_TYPE_EXPERIMENT = "experiment"
 DATA_TYPE_PHENOPACKET = "phenopacket"
+DATA_TYPE_FHIR = "fhir"
```
davidlougheed, line 12: We're still importing "phenopacket" data with FHIR since it gets translated - unless we want to consider tables stored in the FHIR format to be somehow distinct, in which case we can make a fhir app with it's own models

chord_metadata_service/chord/ingest.py
```diff
@@ -62,6 +63,27 @@
                     "value": "{json_document}"
                 }
             ]
+        },
+        "fhir_json": {
+            "name": "FHIR resources json",
+            "description": "This ingestion workflow will validate and import a FHIR schema-compatible "
+                           "JSON document.",
+            "data_type": "experiment",
```
davidlougheed, line 71: data_type should either be fhir or phenopacket (probably latter here, see above)

chord_metadata_service/chord/data_types.py
zxenia: Ok I didn't get this part, so I can remove DATA_TYPE_FHIR at all.

chord_metadata_service/chord/ingest.py
zxenia: I will fix to for 'phenopacket'.

chord_metadata_service/chord/ingest.py
```diff
@@ -247,4 +269,5 @@ def ingest_phenopacket(phenopacket_data, table_id) -> pm.Phenopacket:
 DATA_TYPE_INGEST_FUNCTION_MAP = {
     DATA_TYPE_EXPERIMENT: ingest_experiment,
     DATA_TYPE_PHENOPACKET: ingest_phenopacket,
+    DATA_TYPE_FHIR: ingest_fhir
```
zxenia, line 272: @davidlougheed To what type do I set here 'ingest_fhir' view then? (if I remove fhir data type)

davidlougheed: I'll take a look - I think i'll need to tweak it a bit to check the workflow ID. I'll try to do that before 2pm or so

zxenia: Ok thanks!

PR #144, opened 2020-07-07
Fix deceased

zxenia:


chord_metadata_service/mcode/parse_fhir_mcode.py
```diff
@@ -50,8 +50,13 @@ def patient_to_individual(resource):
     if "active" in resource:
         if resource["active"]:
             individual["active"] = True
-    if "deceasedBoolean" in resource:
+    if "deceasedDateTime" in resource:
         individual["deceased"] = True
+    elif "deceasedBoolean" in resource:
+        if resource["deceasedBoolean"]:
```
davidlougheed, line 56: if deceasedBoolean is a bool then it this could just be
```python
individual["deceased"] = resource["deceasedBoolean"]
```

PR #148, opened 2020-07-13
Update to version and authors

zxenia:
change version to 1.0.0 in package.cfg


chord_metadata_service/package.cfg
```diff
@@ -1,4 +1,4 @@
 [package]
 name = chord_metadata_service
```
davidlougheed, line 2: for v1, are we changing the name or not?

PR #152, opened 2020-08-13
Candig authz

afrosimon:
So we've got:

- A new middleware that will only intervene if a specific env var is set AND for some endpoints (simple list for now)
- As with chord_drs I've setup a dotenv, shouldn't change anything for CHORD and will make it more portable for the future
- Updated the README accordingly

(BTW I've confirmed with Shaikh that this is good to go for CanDIG, regarding the quick discussion we had in today's meeting with the patching of settings.py)


chord_metadata_service/restapi/middleware.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        url = f"{settings.CANDIG_OPA_URL}/v1/data/metadata"
+
+        res = requests.post(url, json=params)
+        res.raise_for_status()
+
+        data = res.json()
+
+        if 'result' in data and 'allow' in data['result']:
```
davidlougheed, line 62: this could be simplified into

```python
return data.get("result", {}).get("allow", False)
```

but maybe this is a bit too code-golfy

chord_metadata_service/restapi/middleware.py
```diff
@@ -0,0 +1,65 @@
...
+            # TODO: unpack the JWT and get access level
+            access_level = 4
+
+            try:
+                allowed = self.query_opa(request, access_level)
+            except requests.exceptions.RequestException:
+                return HttpResponseForbidden()
+
+            if allowed:
+                response = self.get_response(request)
```
davidlougheed, line 29: this could just be an early return

chord_metadata_service/restapi/middleware.py
```diff
@@ -0,0 +1,65 @@
...
+            else:
+                return HttpResponseForbidden()
+        else:
+            response = self.get_response(request)
```
davidlougheed, line 33: this could just be a return

chord_metadata_service/restapi/middleware.py
afrosimon: I've been know to partake in golf once in a while

chord_metadata_service/restapi/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+
+            try:
+                allowed = self.query_opa(request, access_level)
+            except requests.exceptions.RequestException:
+                return HttpResponseForbidden()
+
+            if not allowed:
+                return HttpResponseForbidden()
+
+        response = self.get_response(request)
```
davidlougheed, line 31: What I meant by the early return thing (although it's changed somewhat) is that this could just be
```python
return self.get_response(request)
```
no worries though

PR #164, opened 2020-10-13
Version 1.1.0

zxenia:
1. Improved ingestion error handling
2. Added CanDIG auth middleware
3. Added django-cors
4. Added filtering for phenopackets, patients, resources, experiments apps
5. Changed name to Katsu


chord_metadata_service/chord/ingest.py
```diff
@@ -23,6 +23,7 @@
 __all__ = [
     "METADATA_WORKFLOWS",
     "WORKFLOWS_PATH",
+    "IngestError",
     "ingest_resource",
```
brouillette, line 27: A bit of nitpicking for consistency:
Should these also be in capital letters with underscores between words?

brouillette (commented):
I have yet to use this so I'm unaware of the full scale of the changes, but on a code level, it looks good to me!

chord_metadata_service/chord/ingest.py
zxenia: No, in this case, it is a list that just exposes all different variables and functions that can be imported from here. And they are of different types that follow their own style, e.g. `IngestError`  follows class style in python recommended by PEP8, `ingest_resource ` follows recommended function style.

brouillette: Got it!
Thank you -- approved

brouillette (approved):

brouillette (approved):

PR #172, opened 2020-11-14
Scripts for ingestion and deletion of data

fcoralsasso:
This Pull Request creates the directory `scripts` at the root of the repository and it adds two scripts under it.
1 - A script for data ingestion (`ingest.py`)
2 - A script that automates the process of deleting the data on the tables passed as arguments (`remove_from_db.py`)

### DIFF
--- a/scripts/ingest.py
+++ b/scripts/ingest.py
@@ -44,8 +44,10 @@ def create_project(katsu_server_url, project_title):
     Return the uuid of the newly-created project.
     """
 
-    project_request = {"title": "", "description": "A new project."}
-    project_request["title"] = project_title
+    project_request = {
+        "title": project_title,
+        "description": "A new project."
+    }
 
     try:
         r = requests.post(katsu_server_url + "/api/projects", json=project_request)
@@ -84,8 +86,8 @@ def create_dataset(katsu_server_url, project_uuid, dataset_title):
     Return the uuid of newly-created dataset.
     """
     dataset_request = {
-        "project": "",
-        "title": "",
+        "project": project_uuid,
+        "title": dataset_title,
         "data_use": {
             "consent_code": {
                 "primary_category": {"code": "GRU"},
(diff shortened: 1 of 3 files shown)


scripts/ingest.py
```diff
@@ -0,0 +1,216 @@
...
+            "consent_code": {
+                "primary_category": {"code": "GRU"},
+                "secondary_categories": [{"code": "GSO"}],
+            },
+            "data_use_requirements": [{"code": "COL"}, {"code": "PUB"}],
+        },
+    }
+
+    dataset_request["project"] = project_uuid
+    dataset_request["title"] = dataset_title
```
zxenia, line 99: same as above

scripts/ingest.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    Create a new katsu table.
+
+    Return the uuid of the newly-created table.
+    """
+
+    table_request = {"name": "table1", "data_type": "phenopacket", "dataset": ""}
+
+    table_request["name"] = table_name
+    table_request["dataset"] = dataset_uuid
```
zxenia, line 133: same as above

scripts/ingest.py
```diff
@@ -0,0 +1,216 @@
...
+        "workflow_outputs": {"json_document": ""},
+    }
+
+    private_ingest_request["table_id"] = table_id
+    private_ingest_request["workflow_params"][
+        "phenopackets_json.json_document"
+    ] = phenopackets_json_location
+    private_ingest_request["workflow_outputs"][
+        "json_document"
+    ] = phenopackets_json_location
```
zxenia, line 164: same as above

zxenia (changes requested):
Tested, works great. Thanks a lot for your work!

There is one minor thing to fix in python dictionaries and after this it's ready to go.

zxenia (approved):

PR #183, opened 2021-01-08
Add some new manage.py commands, small performance tweak

davidlougheed:
* Adds commands to create projects/datasets/tables and ingest data in development from the command line to help test stuff out, with documentation in README
* Adds a table of contents to the README
* Slightly speed up listing phenopackets via the API by performing one less SQL query
* Adds example JSON from our synthetic 1000 genomes dataset for easier testing of multiple phenopacket datasets on the fly
* Adds example data use JSON
* Updates some dependencies


zxenia (commented):
Hi @davidlougheed looks good!
Only one thing I'd suggest renaming the django command scripts and remove 'bento' from them because in this case we introduce another name here and it might be confusing for other katsu users. Thanks!

zxenia (approved):
LGTM!

zxenia:
Thank you, @davidlougheed !

PR #205, opened 2021-03-21
DRS and HTTP support for ingestion

davidlougheed:


chord_metadata_service/chord/ingest.py
```diff
@@ -412,54 +427,154 @@ def _get_output_or_raise(workflow_outputs, key):
...
+
+            # If we get here, we have a DRS object we cannot handle; raise an error.
+            raise IngestError(f"Cannot handle DRS object {file_uri}: No file or http access methods")
+
+        elif parsed_file_uri.scheme in (HTTP_URI_SCHEME, HTTPS_URI_SCHEME):
+            yield _workflow_http_download(tmp_dir, file_uri)
+
+        else:
+            # If we get here, we have a scheme we cannot handle; raise an error.
+            raise IngestError(f"Cannot handle workflow output URI scheme: {parsed_file_uri.scheme}")
```
zxenia, line 508: The tests in chord/tests/test_ingest.py fail with this error now because the file path is `example_phenopacket.json` and it has no `parsed_file_uri.scheme`. What do you think about adding another condition here that can handle paths that are just filenames?

zxenia (changes requested):
The tests fail now due to the error I mentioned in the comment above.

chord_metadata_service/chord/ingest.py
davidlougheed: That's a good idea, I'll do that - for some reason I thought this was handled but I guess I didn't do it right, thanks for pointing it out!

zxenia (approved):
Works great, thanks!

PR #251, opened 2021-07-29
Search fixes

davidlougheed:
* Fixes hard-coding of phenopackets in some cases
* Adds mcode search support basics (future-proofing basically)
* Should also speed up experiment search/API results


zxenia (approved):
Tested with queries for experiments and it looks good!

PR #254, opened 2021-07-30
Rename schema IDs, add a bunch of new ones

davidlougheed:
This should hopefully make debugging search in prod a bit more straightforward - it adds nested `$id` fields to all sub-schemas, even stuff like string fields, to not get `N/A` from the search errors and to generally give more precision with schema failures.

It also renames IDs into a consistent and slightly shorter `katsu:[app name]:[schema name]` format, and fields are tagged like `katsu:[app name]:[schema name]:prop_name` or `katsu:[app name]:[schema name]:item:array_obj_prop`.


zxenia (approved):
LGTM

PR #269, opened 2021-09-15
Add readset workflow

zxenia:


chord_metadata_service/chord/tests/test_api_search.py
```diff
@@ -49,7 +49,7 @@ def test_data_type_list(self):
         r = self.client.get(reverse("data-type-list"))
         self.assertEqual(r.status_code, status.HTTP_200_OK)
         c = r.json()
-        self.assertEqual(len(c), 3)
+        self.assertEqual(len(c), 4)
```
davidlougheed, line 52: if there isnt any object on the other end actually stored in katsu, the data type should be excluded from search

davidlougheed (changes requested):

chord_metadata_service/chord/tests/test_api_search.py
davidlougheed: Hmm I think this might cause bugs in the front end though - since it's not a "real" data type (you can't search for objects that fit the schema or anything right?)

davidlougheed: For the purposes of the workflow it's a data type, but since it's just a workaround to ingest files into DRS I think it shouldn't be listed by the service except in the workflow part- at least that's my intuition

zxenia: the search drop-down data types are fetched from this list of data types, right?

davidlougheed: I think so

zxenia: I am wondering if it's the same data types list that also serves the drop-down in Data manager when creating the table - there we would need it to specify the data type for ingestion.

davidlougheed (approved):

zxenia:
One more thing: should the "readset" data type be added to the TABLE_DATA_TYPE_CHOICES - in the front end it's required to select a table and in order to create a table we need to select a data type. But since we won't have a readset in the database, it's unclear to me how it's going to work.

davidlougheed:
One way to maybe do this is to actually have a model + app in Katsu that just contains the DRS URL or something - essentially a shell that isn't searchable in a useful way for now

davidlougheed:
It's essentially like an array of HTSFile I guess but not phenopacket s formatted

davidlougheed:
We may need to add some stuff to the front end to prevent it from showing up manually i guess

zxenia:
We have a model for ExperimentResult in the experiments app that does the same. The filenames (of the readset files) are already captured there that's why I don't want to add another class that basically does the same and stores duplicating info in the database.

davidlougheed:
Oh right. In that case could the readset data type basically just map to a subset of ExperimentResult with that schema etc? It wouldn't duplicate the data but rather filter it to certain types of files (FASTQ BAM SAM CRAM)

zxenia:
Oh I see what you mean, interesting! Let me try to do it here

chord_metadata_service/chord/ingest.py
```diff
@@ -214,6 +216,29 @@
...
+            "name": "Readset",
+            "description": "This workflow will copy readset files over to DRS.",
+            "data_type": DATA_TYPE_READSET,
+            "file": "readset.wdl",
+            "inputs": [
+                {
+                    "id": "readset_files",
+                    "type": "file[]",
+                    "required": True,
+                    "extensions": [".cram", ".bam", ".bigwig"]
```
davidlougheed, line 229: bigwigs are not read files typically I think

zxenia: I see. Basically, we need some name to represent all possible ExperimentResult file types. I will leave it for now and discuss with others what would be the best naming for it.

PR #289, opened 2022-02-04
Merge CanDIG changes upstream

daisieh:
We're trying to synchronize the changes between our forks to be fairly minimal: can you merge these into your fork, or explain why they shouldn't be merged, because they probably are wrong?


zxenia:
Hi @daisieh , I looked through the files, all those changes are already in develop branch and released. I think candig fork needs to be synced with develop and not vice versa since it's the upstream repo.

PR #311, opened 2022-04-08
Export Datasets to cBioPortal format

ppillot:
This PR adds the ability to export a Dataset as a series of text files suitable for ingestion by cBioPortal.

[cBioPortal files format specifications](https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#introduction)
## Changes
- New API endpoint: `/private/export`
-- Mode `PUSH`
-- JSON payload: 
```json
{
    "format": string.         // For now only `cbioportal` supported. Could be expended in the future 
    "object_type": string //"dataset", "project", "table" depending on the format
    "object_id": string.    // e.g. dataset ID
    "output_path": string // optional argument. local path to export the generated files to. If missing a tar file is attached to the HTTP response
}
``` 

## How to test
Two test suites added: 
- `/chord/tests/test_export_cbio.py` tests cBioPortal internal export functions
- `/chord/tests/test_api_export.py` tests export API endpoint (for now only cbioportal format is supported)

## Not part of this PR
The export workflow has not been completed yet. It requires coupling with the Web and the WES services and will be part of the next step of development.

## Varia
The JSON schema for the REST API payload has been added to `chord/schemas.py` for convenience. The ingestion workflow define theirs in bento_lib. Maybe this definition needs to be shared with other services too (Gohan?) which would warrant having it in bento_lib. For now, having it within Katsu, at least makes this code self contained...

The code makes use of a worfklow name (which resolves to `cbioportal`) that was added to the ingest.py module where the list of Workflows is defined. The `MEDATA_WORFLOWS` dictionnary in that file defines an `ingestion` namespace and an empty `analysis` namespace. I've added an `export` namespace. It looks a bit counterintuitive to have data related with export in a file named ingest.py. I suggest we could leave this for later, when the details of the export workflow are laid out (next step of development)


zxenia (changes requested):
Overall, looks good, thank you! I tested the export API locally and it worked as expected.
I requested some changes mainly to the code style.

chord_metadata_service/chord/export_cbio.py
```diff
@@ -0,0 +1,273 @@
...
+    - TMPRSS2_ERG_FUSION_STATUS
+    - ERG_FUSION_ACGH
+    - SERUM_PSA
+    - DRIVER_MUTATIONS
+    """
+
+    samples = []
+    for sample in results:
+        sample_obj = {
+            'individual_id': sample.individual.id,
```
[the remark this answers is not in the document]
ppillot: Thanks, I did not realize this.
It seems to me that in the case of an unidentified individual we should skip the sample altogether when doing the export to cbioportal. Does it make sense?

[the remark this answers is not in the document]
zxenia: Actually, we would always have an individual - but the sample can be assigned to it via the Phenopacket: standard allows doing it in two ways - from Biosample to Individual or linking Biosample to Phenopacket that is linked to an Individual.

chord_metadata_service/chord/export_cbio.py
```diff
@@ -0,0 +1,277 @@
...
+            'id': sample.id
+        }
+        if sample.sampled_tissue:
+            sample_obj['tissue_label'] = sample.sampled_tissue.get('label', '')
+
+        samples.append(sample_obj)
+
+    columns = samples[0].keys()
```
zxenia, line 191: In that case, if all our samples are linked via Phenopackets (and for all of them `sample.individual` is None) then our `samples` array will be empty. The line 191 `columns = samples[0].keys()` would give us  `list index out of range` error because the array is empty.
I suggest two options here:
1. Do not skip sample and collect its id
2. Get information about patient id via phenopacket.

zxenia (changes requested):

chord_metadata_service/chord/export_cbio.py
```diff
@@ -0,0 +1,277 @@
...
+        # Note: plural in `phenopackets` is intentional (related_name property in model)
+        indiv = Individual.objects.filter(phenopackets__table__ownership_record__dataset_id=dataset.identifier)
+        individual_export(indiv, file_patient)
+
+    with open(getPath(PATIENT_META_FILENAME), 'w') as file_patient_meta:
+        clinical_meta_export(cbio_study_id, PATIENT_DATATYPE, file_patient_meta)
+
+    # Export samples
+    with open(getPath(SAMPLE_DATA_FILENAME), 'w') as file_sample:
+        sampl = pm.Biosample.objects.filter(phenopacket__table__ownership_record__dataset_id=dataset.identifier)
```
zxenia, line 59: maybe here you can use [prefetch_related()](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#prefetch-related) and add information about patient id in the final queryset that is later passed to `sample_export()`

chord_metadata_service/chord/export_cbio.py
ppillot: Done. The code now requests the individual id from phenopacket when it is missing from the `individual` field.
Thanks

chord_metadata_service/chord/export_cbio.py
ppillot: Thanks for the pointer to `prefetch_related`. What I did is adding a test inside a loop over the biosamples.
I am not sure about which one is better: Django documentation states that prefetch_related implies that the whole content of the Biosample table will have to be dumped in memory so that the "related" queries can be prefetched. On the other hand, querying the DB in a loop (albeit conditionally in that case) is not a great option for performance.
Maybe this could be addressed later, when we want to export more data from the biosample: prefetch_related might become unavoidable then.

zxenia (approved):

PR #319, opened 2022-05-16
Add Experiment results endpoint

zxenia:
This PR:
- Adds `api/experimentresults` endpoint to list all available experiment results.
- Adds the following filters on top of the `api/experimentresults`:
`identifier`, `description`, `filename`, `genome_assembly_id`, `file_format`, `data_output_type`, `usage`, `created_by`, `extra_properties`, `datasets`. All filters can be combined with each other.
- Adds `datasets` filter to `api/experiments`.

Note: The `datasets` filter is filtering by a dataset title and can take a list of datasets separated by comma,
e.g. `?datasets=dataset_1,dataset_2`

### DIFF
--- a/chord_metadata_service/experiments/tests/test_api.py
+++ b/chord_metadata_service/experiments/tests/test_api.py
@@ -1,0 +1,133 @@
+import os
+from uuid import uuid4
+
+from rest_framework import status
+from rest_framework.test import APITestCase
+from chord_metadata_service.chord.models import Project, Dataset, TableOwnership, Table
+from chord_metadata_service.chord.tests.constants import VALID_DATA_USE_1
+from chord_metadata_service.chord.data_types import DATA_TYPE_PHENOPACKET, DATA_TYPE_EXPERIMENT
+from chord_metadata_service.chord.ingest import (
+    WORKFLOW_INGEST_FUNCTION_MAP,
+    WORKFLOW_PHENOPACKETS_JSON,
+    WORKFLOW_EXPERIMENTS_JSON,
+)
+
+
+EXAMPLE_INGEST_OUTPUTS_PHENOPACKETS_JSON = {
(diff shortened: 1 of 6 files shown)


chord_metadata_service/experiments/tests/test_api.py
```diff
@@ -0,0 +1,133 @@
...
+EXAMPLE_INGEST_OUTPUTS_PHENOPACKETS_JSON = {
+    "json_document": os.path.join(os.path.dirname(__file__), "example_phenopackets.json"),
+}
+
+EXAMPLE_INGEST_OUTPUTS_EXPERIMENTS_JSON = {
+    "json_document": os.path.join(os.path.dirname(__file__), "example_experiments.json"),
+}
+
+
+class GetExperimentsAppApisTest(APITestCase):
```
ppillot, line 25: Very minor: I don't understand what "App" means in this context (i.e. "ExperimentsApp", "Experiments app APIs")
I would think it's question of experiment results in this test, but I may miss something lexic wise?

ppillot (approved):
This looks good.
Thanks!

ppillot:
After testing, I've noticed that the search on the `datasets` name is case sensitive, whereas the search on the file type is not ("VCF" or "vcf" both return records of type "VCF").
Also, is it possible that a dataset name would contain a comma which could break the query string parsing? I guess that, as the dataset name is used elsewhere in queries (in other API endpoints), it might just be something to enforce when the dataset name is added to the DB (which might already be the case, I haven't checked the code)

chord_metadata_service/experiments/tests/test_api.py
ppillot: I am partially replying to myself: these tests are not only about experimental results, but also experiments.

zxenia:
@ppillot It's possible and there is no check for it. So far this feature was not used much and we are the once who curate the dataset metadata. But yeah we have to think about something here.

PR #321, opened 2022-05-30
Add Controlled vocabularies to Experiments app

zxenia:
This PR adds controlled vocabularies to Experiments app schemas. Since these values will be changing in the next months we only add them to the schemas which are used for data validation during ingestion.


chord_metadata_service/experiments/schemas.py
```diff
@@ -84,27 +85,36 @@
             "type": "string"
         },
         "study_type": {
-            "type": "string"
+            "type": "string",
+            "enum": ["Genomics", "Epigenomics", "Metagenomics", "Transcriptomics", "Other"]
```
ppillot, line 89: I am wondering about the capitalization here. Is it of no importance or is this schema enforcing the capitalization for this property?

chord_metadata_service/experiments/schemas.py
```diff
@@ -84,27 +85,36 @@
...
-            "type": "string"
+            "type": "string",
+            "enum": ["DNA Methylation", "mRNA-Seq", "smRNA-Seq", "RNA-Seq", "WES", "Other"]
         },
         "experiment_ontology": ONTOLOGY_CLASS_LIST,
         "molecule": {
-            "type": "string"
+            "type": "string",
+            "enum": ["total RNA", "polyA RNA", "cytoplasmic RNA", "nuclear RNA",
+                     "small RNA", "genomic DNA", "protein", "Other"]
```
ppillot, line 99: Here, the adjectives are not capitalized, `Other` is, but `protein` is not.

chord_metadata_service/experiments/schemas.py
```diff
@@ -28,15 +28,16 @@
...
                      "FASTQ", "TAB", "SRA", "SRF", "SFF", "GFF", "TABIX", "UNKNOWN", "OTHER"]
         },
         "data_output_type": {
             "type": "string",
             "enum": ["Raw data", "Derived data"]
         },
         "usage": {
-            "type": "string"
+            "type": "string",
+            "enum": ["visualized", "downloaded"]
```
ppillot, line 40: No capitalization for these values (because they are adjectives? But note that `library_layout` values are capitalized (`Single`, `Paired`)

ppillot (approved):
All looks good.
There is a possible issue with capitalization consistency. It might be of no consequence here, but I am no expert in the matter.

chord_metadata_service/experiments/schemas.py
zxenia: Capitalized them now, thanks for catching this error!

PR #327, opened 2022-06-30
Features/fast overview

ppillot:
This branch changes the inner workings of the overview REST API endpoint by using DB queries to compute statistics (previously it was doing a single query with multiple joins and computing statistics by parsing the tables of results).

**Intentional change**: the statistics about age are returned already binned. This can be reverted if deemed needed.

**How to test:**
- ensure that the previous response and this one are identical. This can be done by loading the overview page in a browser and inspecting the network

**Known issue:**
~~In some cases the `age` property can be a range. This is not handled by the current code.~~

### DIFF
--- a/chord_metadata_service/restapi/api_views.py
+++ b/chord_metadata_service/restapi/api_views.py
@@ -4,11 +4,11 @@ import logging
 from collections import Counter
 from django.conf import settings
 from django.views.decorators.cache import cache_page
+from django.db.models import Count
 from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.decorators import api_view, permission_classes
 
-from chord_metadata_service.phenopackets.api_views import PHENOPACKET_PREFETCH, PHENOPACKET_SELECT_REL
 from chord_metadata_service.restapi.utils import parse_individual_age
 from chord_metadata_service.chord.permissions import OverrideOrSuperUserOnly
 from chord_metadata_service.metadata.service_info import SERVICE_INFO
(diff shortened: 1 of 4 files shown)


ppillot:
Age defined as ranges are handled in the latest iteration

gsfk:
Nice! This is _much_ faster, where the old version was already too slow. 

This version bins ages by decade, where the previous version binned by year, which required an extra step of binning by decade in the front end. I think the new approach is preferable. 

Still a couple bugs to work out, though, since local tests show a few quirks: 
- all diseases have a count of one
- all instrument platforms have a count of one
- sampled tissue counts disagree with counts given by the current katsu verison (and for the handful that I checked the current version was correct)

gsfk (commented):
Nice! This is _much_ faster, where the old version was already too slow. 

This version bins ages by decade, where the previous version binned by year, which required an extra step of binning by decade in the front end. I think the new approach is preferable. 

Still a couple bugs to work out, though, since local tests show a few quirks: 
- all diseases have a count of one
- all instrument platforms have a count of one
- sampled tissue counts disagree with counts given by the current katsu verison (and for the handful that I checked the current version was correct)

ppillot:
@gsfk Thanks for your tests! These highlight issues with the synthetic dataset as well: there is only one disease referenced for only one patient...
The disease and the instruments related stats have been fixed now.
Regarding the Biosamples, let's discuss it: I currently don't understand where is the issue... there is something I probably miss in my comprehension of the relationship between phenopackets and biosamples.

chord_metadata_service/restapi/api_views.py
```diff
@@ -44,173 +47,87 @@ def overview(_request):
...
-                "model": dict(instruments["model"])
+                "count": instruments_count,
+                "platform": stats_for_field(experiments_models.Experiment, "instrument__platform"),
+                "model": stats_for_field(experiments_models.Experiment, "instrument__model")
             },
         }
-    })
+    }
+
+    return Response(r)
```
SanjeevLakhwani, line 130: Just wondering, the motivation for changing it to a variable instead of a direct response

chord_metadata_service/restapi/api_views.py
```diff
@@ -452,3 +369,48 @@ def sort_numeric_values_into_bins(values: dict, bin_size: int = 10, threshold: i
...
+
+def compute_binned_ages(bin_size):
+    a = pheno_models.Individual.objects.filter(age_numeric__isnull=True).values('age')
+    binned_ages = []
+    for r in a.iterator():  # reduce memory footprint (no caching)
+        if r["age"] is None:
+            continue
+        age = parse_individual_age(r["age"])
+        binned_ages.append(age - age % OVERVIEW_AGE_BIN_SIZE)
+    return binned_ages
```
SanjeevLakhwani, line 416: Might be better to have them nested within the response function, or have a utility function set file for it, as then the file would have a better logical arrangement that each function as an api response function and not an util function

SanjeevLakhwani (commented):
Looks great, I do not have anything to add on but just 2 discussions. Great work! and Love the changes. I was also wondering if using numpy would be an option for the project as that would have faster mathematical calculation speeds for functions such as compute_binned_ages and possibly at other places too.

chord_metadata_service/restapi/api_views.py
ppillot: Easier to debug by placing breakpoints. Can be simplified.

PR #330, opened 2022-07-26
Features/public chart config

ppillot:
This PR implements a new config file for the public entry point of Katsu that supersedes the previous version.
The aim is to decouple the config file structure and response from the DB organization (in the previous version, there was a nesting for the fields defined in extra properties), clarify which part of the config file refers to distinct entities (e.g. datasets, counts are distinct from discrete fields) and avoid processing on the client side whenever it can be done on the backend.

As part of this code review, consider inspecting the chord_metadata_service/example.config.json file and reading the updated documentation file for the public API.

### Settings:
Make sure to update the `config.json` file with the content of the `example.config.json` file
If you don't have a suitable dataset to base your tests upon, you can generate an updated bento demo dataset from the relevant project (updated recently to have examples of all field types) or download an already generated version from the datateam gdrive

### To test:
- api/public_overview: it returns the statistics per field. Check that the data part of the response is consistent: with the config file (enum values are enforced for `string` fields, binning is obeyed for `number` fields, date range is consistent for `date` fields), and with the DB content
- api/public_search_fields: based on the config file, it returns only the fields defined in `config.search`, ordered by sections. Of note are the `options`, which are the values that can be used in the query. Should be consistent with the config file.
- api/public: it accepts a combination of field names and values such as defined in the previous endpoint. `api/public?age=30-40&date_of_consent=Mar 2021`


ppillot:
Code has been inspected by @gsfk 
Test revealed an issue when dev container was used in conjonction with the latest public UI. This does not seem to be related with the code in this PR, but an issue with concurrency in this particular setup.
Some mitigation has been put in place in the bento-public frontend to alleviate this issue.

PR #333, opened 2022-08-18
Features/leaner queries

ppillot:
This PR implements new features to improve the performance of the "federated" search in Bento.

# Context of the PR
The federation service hits Katsu on the /private/[table-id]/search/ url with a query (either POST or GET) containing a `query` parameter which is a string in the format of bento-lib free form queries. The `table-id` extracted from the URL is used to fetch the table properties such as its `data-type` (i.e. "phenopacket", "mcodepacket", "experiment")

Within katsu, the `query` parameter is converted to an SQL statement using bento-lib and psycopg2. This returns a list of IDs for the given table `data_type`. This list of ids is used in a second query made via the Django ORM to the model corresponding to the `data_type` (for example, search for all phenopackets matching this list of IDs)

In general, the ORM also prefetches related records from the accompagnying tables (for examples Biosamples related with the list of phenopackets).

The last step is the serialization of the results from the queries. This makes use of serializers defined for the different Katsu applications (i.e. the serializer for phenopackets also calls the serializers for Patients and for Biosamples,... when traversing the list of results). This last step takes the longest time and produces a bloated response.

# New features
The federation service aggregates results using a "linked field" that is defined at the level of the Dataset object and is common between all the queried data_types. The main idea here is to reduce the amount of data transferred just to the list of ids matching the query for the "linked field".

To do that 2 optional parameters are added to this PR:
- `output` which describes a format expected for the response. In the previous case it is `values_list` which would return a result containing an array of values (not a nested object)
- `field` which is the field to use for fetching this value. The field syntax is given using the bento-lib syntax, which is also the syntax used for the Dataset/linked_field_sets property

In a first attempt I've tried to do without the `field` property, and trying to extract it directly from the dataset table based on the table data_type. Even if the code was working, this could have lead to unexpected results from the requester, for example in the case when multiple sets of linked fields had been defined. Having `field` as a parameter provided by the sender simplifies the code and ensures that the service that asks for the data is responsible for defining what it requires.

Note also that I've tried to name the `output` parameter `format` in a first place. This worked well in a POST request, but caused issues when doing a GET request, due to Django REST Framework that tries to automatically find a Renderer class corresponding to this format. It would have required substantial development to make this work in the broader context of DRF, both for GET and POST queries which is why the parameter name `output` has been chosen instead.

The second feature applies to the case of bento-web interface where only a subset of the values from the phenopackets is displayed to the user (subject id, list of matching biosamples, number of related experiments). In a second step of the search by the federation service, once all the results have been collated and intersected, a new query containing the intersection of list of ids is sent, with the `output` parameter set to `bento_search_result`. The `field` parameter does not apply in this case.

## How to test
- The general test suite must pass
- using the swagger API, try to reach the /private/{table_id}/search endpoint and run queries with the given `output` and `field` parameters.

### DIFF
--- a/chord_metadata_service/chord/tests/constants.py
+++ b/chord_metadata_service/chord/tests/constants.py
@@ -20,6 +20,7 @@ __all__ = [
     "TEST_SEARCH_QUERY_7",
     "TEST_SEARCH_QUERY_8",
     "TEST_SEARCH_QUERY_9",
+    "TEST_SEARCH_QUERY_10",
     "TEST_FHIR_SEARCH_QUERY",
 ]
 
@@ -249,4 +250,6 @@ TEST_SEARCH_QUERY_6 = ["#ico", ["#resolve", "biosamples", "[item]", "sampled_tis
 TEST_SEARCH_QUERY_7 = ["#eq", ["#resolve", "experiment_results", "[item]", "file_format"], "VCF"]
 TEST_SEARCH_QUERY_8 = ["#ico", ["#resolve", "experiment_type"], "chromatin"]
 TEST_SEARCH_QUERY_9 = ["#eq", ["#resolve", "subject", "id"], "patient:1"]
+TEST_SEARCH_QUERY_10 = ["#in", ["#resolve", "biosamples", "[item]", "id"],
+                        ["#list", "biosample_id:1", "biosample_id:2"]]
 TEST_FHIR_SEARCH_QUERY = {"query": {"match": {"gender": "FEMALE"}}}
--- a/chord_metadata_service/chord/tests/test_api_search.py
+++ b/chord_metadata_service/chord/tests/test_api_search.py
@@ -515,7 +515,7 @@ class SearchTest(APITestCase):
 
     def test_private_table_search_bento_search_results(self):
         # Valid query to search for biosample id in list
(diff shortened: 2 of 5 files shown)


gsfk (approved):
Great! Clear and easy to use, and integrates well with beacon.

Only very minor comments for the "bento_search_result" option: 
- `table_id` is repeated for every result, is this necessary? Is table_id needed at all?
- `alternate_ids` is listed even when null

Both of these add to the size of the response, although that seems like a quibble when the changes here already make significant reductions in the response size.

ppillot:
Thanks!
Your comments are point-on.
- `table_id` was added in the last commits to allow the grouping by table_id when doing a search by contrast with a table search. I agree, that as it is not used by Bento, it should be stripped out from the Bento Search Response.
- `alternate_ids` is a value used by bento_web. I think that the response should be consistent and then all the attributes shall be present which has the advantage of reducing the code necessary for handling the edge cases in the client that consumes the API response. Nonetheless, when there are no `alternate_ids`, at the moment a `null` value is returned when we are expecting an array. I think it should be an empty array instead.
I'll add these two modifications. Thanks for pointing that out!

gsfk:
Changes look good. The extra handling that checks for alternate_ids is already present in the front end code, but there are also merits in a consistent response.

PR #335, opened 2022-09-02
Features/serve dats file

ppillot:
New feature: endpoint `public_dataset` added which returns a JSON response as
```json
{
    "datasets": [
        {
            "title": ...
        },
        ...
    ]
}
```
New feature: field dats_file added to the dataset model

### How to test
You need to run the migrations with the command `python manage.py migrate`
This will create the new field in the DB

To load a JSON dats file, you would need to use the Django shell, import the datasets model and update your current model
```python
python manage.py shell
from chord_metadata_service.chord.models import Dataset
with open('chord_metadata_service/restapi/tests/example_dats_provenance.json') as f: dats = f.read()
d = Dataset.objects.all()[0]
d.dats_file = dats
d.save()
exit()
```

### What to test
New endpoint at `api/metadata/public_dataset`
Changed endpoint at `api/metadata/public_overview` must not contain the `datasets` property anymore.


SanjeevLakhwani (commented):
Great work. I was wondering if we would be able to save provenance data as a file itself without using of a db to possible have less processing required to return a response.

ppillot:
That's an option I did consider. But to do that, we'd need to have a place to store assets which is not the case in Katsu currently. Also these assets could be updated, and need to be deleted when a Dataset is removed. The naming of the file must be handled (even if it's trivial, it's another thing to think about).
On the other hand, storing in the database, with the Dataset model, offers all of these functionalities for free. And adding a UI to create/update the DATS file will be trivial (no need of another API endpoint, just another input in the dataset form).
A common issue with storing files in the DB is that LOBs (Large Objects) can be notoriously slow to handle when performing queries. I had the experience of a DB with such large files (around 1MB) and fetching records was taking a noticeable time. We had to select the fields we needed (which were not the LOBs) to reduce the time for processing and in some cases store the lobs on disk instead of in DB.
In the case of Katsu Dataset, there is at most of couple of Datasets and the JSON file is quite short (fits in a long text) so it's still an implementation that can scale without much care to dedicate for the edge cases.

PR #339, opened 2022-09-21
Features/stats list

ppillot:
This branch adds some statistics on filtered results for certain fields. This functionality is used for 2 new features, which are grouped in this branch:
- new endpoint: api/search_overview. This endpoint receives a POST request with the following payload:
```json
{
    "id": ["patient:1", "patient:2",...]
}
```
and it returns an object with various statistics (counts of biosamples and experiments for this subset, detailed aggregated statistics for some fields (sampled_tissue, histological_diagnosis...). See the following ticket for requirements: https://206.12.92.46/issues/1258
The goal is to provide an overview of search results in the Bento UI.
- api/public endpoint: previously was returning only a count of results. This commit adds statistics about sampled_tissue and experiment_type. See the following ticket for requirements https://206.12.92.46/issues/1211

How to test:
The test suite has been updated in regard with those changes
- for the public interface, one can perform a search such as `api/public?sex=male`
- for the search_overview, using the demo dataset I've tried the following `api/search_overview?id=ind:NA19648&id=ind:HG00096`
ideally this should be tested in the context of bento_web when the implementation of fetching search results is completed.

### DIFF
--- a/chord_metadata_service/restapi/api_views.py
+++ b/chord_metadata_service/restapi/api_views.py
@@ -13,6 +13,7 @@ from chord_metadata_service.restapi.utils import (
     get_field_options,
     parse_individual_age,
     stats_for_field,
+    queryset_stats_for_field,
     compute_binned_ages,
     get_field_bins,
     get_categorical_stats,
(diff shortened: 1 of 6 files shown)


chord_metadata_service/restapi/api_views.py
```diff
@@ -138,6 +129,58 @@ def overview(_request):
...
+@api_view(["GET", "POST"])
+@permission_classes([OverrideOrSuperUserOnly])
+def search_overview(request):
+    """
+    get+post:
+    Overview statistics of a list of patients (associated with a search result)
+    - Parameter
+        - id: a list of patient ids
+    """
+    individual_id = request.GET.getlist("id") if request.method == "GET" else request.data.get("id", [])
```
SanjeevLakhwani, line 141: As it is an array that can contain multiple values, might be better to name individual_id as individual_ids

SanjeevLakhwani (commented):
Great job, everything seems to be good!

chord_metadata_service/restapi/api_views.py
ppillot: Yes! Good suggestion. I'll add it in a subsequent PR.
Thanks for your review!

PR #342, opened 2022-10-04
Features/fast text search

ppillot:
This PR speeds up the keyword based search in Bento by skipping the usage of the IndividualSerializer

## Some context about DRF ViewSets
A ViewSet in Django Rest Framework abstracts all the operations related to the controller part.
It defines a couple of methods such as `details` which is called with URLS of the form `/api/individuals/indiv:id` with a GET method to output the details about a single record. The `list` method which is overridden in this PR, is called by a GET request on an url of the form `api/individuals/`.
This class defines also some properties such as:
- serializer_class: a class that receives a queryset and converts it to a python object
- renderer_classes: a list of classes that receive a serialized object and convert it to an http response
- filter_backends: a list of filter objects which receive a queryset and a request object, and apply some filtering to the queryset based on the parameters of the request
- `filterset_class`: comes from the Django_filter library and is used to facilitate the definition of filters based on request parameters. This is where the logic for the fulltext search is defined
- queryset: a Django query definition

## Code logic
in this PR, the `list` method is overridden to test for the format used as a parameter. In case of bento_search_results, the `queryset` is manually managed to apply the filtering using the defined `filterset_class` and query for related phenopackets using the expected fields for the bento search response. At the end of the method a Response is returned which takes directly the list of results from the queryset, instead of going through serialization.

## To test
The test suite should work
The PR can be tested in the context of Bento. None of the other search functionality shall be broken.

### DIFF
--- a/chord_metadata_service/patients/api_views.py
+++ b/chord_metadata_service/patients/api_views.py
@@ -115,8 +115,8 @@ class IndividualBatchViewSet(BatchViewSet):
     content_negotiation_class = FormatInPostContentNegotiation
 
     def get_queryset(self):
-        individual_id = self.request.data.get("id", None)
-        filter_by_id = {"id__in": individual_id} if individual_id else {}
+        individual_ids = self.request.data.get("id", None)
+        filter_by_id = {"id__in": individual_ids} if individual_ids else {}
         queryset = Individual.objects.filter(**filter_by_id)\
             .prefetch_related(
                 *(f"phenopackets__{p}" for p in PHENOPACKET_PREFETCH if p != "subject"),
--- a/chord_metadata_service/patients/api_views.py
+++ b/chord_metadata_service/patients/api_views.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from rest_framework import viewsets, filters, mixins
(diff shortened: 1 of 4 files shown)


gsfk (approved):
Thanks for this! Searches are much faster, in most cases several times faster than existing code. Performance is still slow though, all searches take roughly the same time (2.5 min) regardless of the search:

```
TERM 					OLD 		THIS PR
"I have no problems in walking about"	11 min		2.6 min
"Hospitalized adult"			9.1 min		2.5 min
headache				8.8 min		2.5 min
(a specific sample id)			3.9 min		2.5 min
```

(Times are with debugpy running). Is the time now dominated by the time needed to flatten text?

ppillot:
Thanks for your testings. After discussion, it appears clearly that the full text search takes a long time on the BQC19 instance, presumably due to the amount of extra-properties. This PR only addresses the serialization part of the search which was taking a lot of time.
To further improve the text search speed, one should inspect the query plan from postgresql to have insights at where the slowdowns are located. It's possible some indexing could be added. Another option is to add a table for the full text search purpose by laying out the required properties.

PR #343, opened 2022-10-04
Fix/table summary

ppillot:
This PR fixes the issue described in the ticket https://206.12.92.46/issues/1038 where the counts for sex are incorrect (doubled).
The implementation relies on the same method as for the overview summary, public overview and search results statistics, which improves the response speed (from 20s to 200ms on the chord_demo dataset).
The age stats have also been added to the response as a comment let think it was the original intent.

## How to test:
- In the context of Bento, use Admin>Data Manager, click on the [Data Tables] tab and find a phenopacket table in the list. Click on the blue link with the table UUID in the first column.
- From Katsu, the endpoint is located at /api/metadata/api/tables/56b44259-1fff-46ef-9f0d-5897fea63eee/summary (where the UUID is the table UUID) and can be checked using the swagger interface at /api/metadata/

### DIFF
--- a/chord_metadata_service/chord/views_search.py
+++ b/chord_metadata_service/chord/views_search.py
@@ -179,30 +179,30 @@ def mcodepacket_table_summary(table):
 
 
 def phenopacket_table_summary(table):
-    phenopackets = Phenopacket.objects.filter(table=table)  # TODO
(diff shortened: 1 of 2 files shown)


brouillette:
image
LGTM!

PR #344, opened 2022-10-06
Features/public config custom

ppillot:
## New features
- allow custom bins definition in public config for numeric fields
- allow different field mappings for overview (statistics) and search (in current Bento Public, searches return a number of individuals, so fields must be defined in relation to the Individual model)

## How to test
- Test suite should run without error
-  The example of config file has been updated in ./chord_metadata_service/example.config.json. In the context of Bento, this file can be copied in bentov2/lib/katsu/config.json which is automatically mounted in ./chord_metadata_service/config.json in the bentov2-katsu container. The container must be restarted for the changes to take properly effect. From there, it should be possible to see the expected results in Bento_public for the fields lab_test_results and experiment_types.

### DIFF
--- a/chord_metadata_service/example.config.json
+++ b/chord_metadata_service/example.config.json
@@ -98,11 +98,8 @@
             "description": "This acts as a placeholder for numeric values",
             "datatype": "number",
             "config": {
-                "bin_size": 50,
-                "taper_left": 50,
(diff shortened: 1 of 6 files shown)


brouillette:
image
Running locally, and the configuration populates to the bento-public front-end -- LGTM!

PR #345, opened 2022-10-06
Features/beacon search latest

gsfk:
Adds a new endpoint `/api/beacon-search` for searches forwarded from [beacon](https://github.com/bento-platform/bento_beacon). 

Beacon is a data discovery service focusing on genetic variants. [Version 2 of beacon](https://beacon-project.io/) has the ability to filter queries by clinical or other metadata, so needs the ability to query both katsu and gohan. Note that this endpoint is protected by the gateway only, similar to `/api/individuals` and other endpoints. 

Also in this pr:
- tweaked pagination limits for katsu autocomplete endpoints, which are called by beacon
- tests

### DIFF
--- a/chord_metadata_service/phenopackets/autocomplete_views.py
+++ b/chord_metadata_service/phenopackets/autocomplete_views.py
@@ -3,7 +3,7 @@ from .models import Disease, PhenotypicFeature, Biosample
 
 
 class DiseaseTermAutocomplete(autocomplete.Select2QuerySetView):
-    paginate_by = 50
+    paginate_by = 200
 
     # get_result_value return result.pk
 
@@ -27,7 +27,7 @@ class DiseaseTermAutocomplete(autocomplete.Select2QuerySetView):
 
 
 class PhenotypicFeatureTypeAutocomplete(autocomplete.Select2QuerySetView):
-    paginate_by = 50
+    paginate_by = 200
 
     def get_result_value(self, result):
         # returns phenotypic feature type ontology id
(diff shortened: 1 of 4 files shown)


chord_metadata_service/phenopackets/autocomplete_views.py
```diff
@@ -3,7 +3,7 @@
 
 
 class DiseaseTermAutocomplete(autocomplete.Select2QuerySetView):
-    paginate_by = 50
+    paginate_by = 200
```
ppillot, line 6: Consider using a constant defined at the beginning of the file instead of copying a number at 3 different places.

chord_metadata_service/phenopackets/autocomplete_views.py
```diff
@@ -27,7 +27,7 @@ def get_queryset(self):
 
 
 class PhenotypicFeatureTypeAutocomplete(autocomplete.Select2QuerySetView):
-    paginate_by = 50
+    paginate_by = 200
```
ppillot, line 30: Same as above

chord_metadata_service/phenopackets/autocomplete_views.py
```diff
@@ -46,7 +46,7 @@ def get_queryset(self):
 
 
 class BiosampleSampledTissueAutocomplete(autocomplete.Select2QuerySetView):
-    paginate_by = 50
+    paginate_by = 200
```
ppillot, line 49: Same as above

ppillot (changes requested):
See the nitpicking comment about using constant for the repeated page_size value.
Otherwise, I am not sure that a dedicated endpoint is required for beacon. I think beacon could reach Katsu on the existing private/search endpoint by using the docker network (a url such as http://bentov2-katsu:8000) as the federation service does as it appears it's the same usage.

gsfk:
Closing as unnecessary. I needed this new endpoint for local non-container dev only. Not needed for prod or containerized dev, so no need to merge. I'll make a fresh PR for the pagination limit changes.

PR #348, opened 2022-10-12
fix for biosamples overcount in search results

gsfk:
Fixes two issues with biosamples in Bento search results:

- biosamples were overcounted when multiple experiments were performed on a single sample (Redmine [#1298](https://206.12.92.46/issues/1298))
- individuals with no biosamples were mistakenly listed as having one biosample (a "null" biosample).

The same fix appears twice (there is distinct code for Bento "Advanced Search" and "Text Search").


davidlougheed (approved):

davidlougheed:
lgtm

PR #350, opened 2022-10-26
remove enum from experiments usage property

gsfk:
Remove controlled vocabulary from experiments results "usage" field.


brouillette (approved):
LGTM!

davidlougheed:
do we need `usage` at all?

gsfk:
Possibly not! The current plan seems to be:
- remove controlled vocabulary for current Bento release (2.8) so that this field is no longer validated
- revisit usefulness in next release

PR #352, opened 2022-11-04
Bugfix/bento public missing

gsfk:
This fixes several issues with the response from `api/public` and `api/public_overview` used by bento_public.

Previously, `api/public` returned only a count of individuals. The current version returns a richer results set, including; 
- count of biosamples
- count of experiments 
- array of counts for `sampled_tissue` in biosamples
- array of counts for `experiment_type` in experiment 

In many cases, some or all of these values were incorrect. Changes include: 

- removing censorship in counts for `sampled_tissue` and `experiment_type` in searches. Censorship of searches is all-or-nothing by count of individuals, not line-by-line in breakdown of statistics. (However, censorship still remains at the category level for overview statistics). 
- correcting the handling of missing values in querysets (this affects both overview and search)
- hardcoding queries for `sampled_tissue` and `experiment_type` to handle the statistics returned by `api/public`.
- fix for off-by-one error in generation of month labels (Dec 2021 was placed immediately before January 2021)

### DIFF
--- a/chord_metadata_service/restapi/utils.py
+++ b/chord_metadata_service/restapi/utils.py
@@ -227,7 +227,7 @@ def monthly_generator(start: str, end: str) -> Tuple[int, int]:
     [end_year, end_month] = [int(k) for k in end.split("-")]
     last_month_nb = (end_year - start_year) * 12 + end_month
     for month_nb in range(start_month, last_month_nb):
-        year = start_year + month_nb // 12
+        year = start_year + (month_nb - 1) // 12
         month = month_nb % 12 or 12
         yield year, month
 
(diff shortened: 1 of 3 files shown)


chord_metadata_service/restapi/utils.py
```diff
@@ -227,7 +227,7 @@ def monthly_generator(start: str, end: str) -> Tuple[int, int]:
     [end_year, end_month] = [int(k) for k in end.split("-")]
     last_month_nb = (end_year - start_year) * 12 + end_month
     for month_nb in range(start_month, last_month_nb):
```
davidlougheed: range is semi-open: `[start_month, last_month_nb)`: does this work as intended right now? because elsewhere where this is used seems to imply that it should be inclusive of end, but it hasn't been so far...

chord_metadata_service/restapi/utils.py
```diff
@@ -506,11 +486,11 @@ def get_range_stats(field_props):
...
-                  **{f"{field}__gte": floor} if floor is not None else {},
-                  **{f"{field}__lt": ceil} if ceil is not None else {},
-                  then=Value(label)
-                 )
-             for floor, ceil, label in labelled_range_generator(field_props)]
+        **{f"{field}__gte": floor} if floor is not None else {},
+        **{f"{field}__lt": ceil} if ceil is not None else {},
+        then=Value(label)
+    )
+        for floor, ceil, label in labelled_range_generator(field_props)]
```
davidlougheed: slightly awkward formatting, maybe combine these 2 lines `) for floor, ...`

davidlougheed (commented):
overall looks good, just one question about `range()`

chord_metadata_service/restapi/utils.py
gsfk: Yes, nice catch. I already fixed an off-by-one error in here, but you're right that this misses the last month also.

davidlougheed (approved):

davidlougheed:
lgtm

chord_metadata_service/restapi/utils.py
gsfk: My only contribution here is accidental whitespace changes (which VS Code seems to hide by default now? That's not a feature I wanted). 

Honestly, I'm expecting to refactor much of this file in the near future, if only to pull out the code that collects counts for "missing".

PR #355, opened 2022-11-09
remove test

gsfk:
Remove iffy test that fails for unclear reasons. Covers the same code as  `test_public_get()` so not crucial.


davidlougheed (approved):
LGTM

PR #361, opened 2022-11-18
fix issue with 0-age individuals, use interval notation for public endpoints, and clean up restapi.utils

davidlougheed:


gsfk (approved):
Looks good

PR #362, opened 2022-11-23
add git info to katsu service-info

noctillion:
Add git and environment info to service-info in katsu


davidlougheed:
for the PR actions to run, you'll need to fix the merge conflict btw

PR #363, opened 2022-11-25
add biosamples count to public_overview response

gsfk:
- add biosamples count to public_overview response

### DIFF
--- a/chord_metadata_service/restapi/api_views.py
+++ b/chord_metadata_service/restapi/api_views.py
@@ -352,6 +352,7 @@ def public_overview(_request):
 
     # Predefined counts
     individuals_count = patients_models.Individual.objects.all().count()
+    biosamples_count = pheno_models.Biosample.objects.all().count()
     experiments_count = experiments_models.Experiment.objects.all().count()
 
     # Early return when there is not enough data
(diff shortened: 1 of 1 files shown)


chord_metadata_service/restapi/api_views.py
```diff
@@ -352,6 +352,7 @@ def public_overview(_request):
 
     # Predefined counts
     individuals_count = patients_models.Individual.objects.all().count()
+    biosamples_count = pheno_models.Biosample.objects.all().count()
```
davidlougheed, line 355: raising an "early" warning that at some point this all will break / be incorrect if more than one dataset is in an instance, since these aren't dataset specific (and should be, since the rest of Bento Public seems to present a specific dataset) – we should either remove the concept of datasets/projects completely or follow the structure

davidlougheed (approved):
looks good to me

noctillion (approved):
Looks good!!

chord_metadata_service/restapi/api_views.py
gsfk: Yeah, this is a good point. For bento_public it's not even clear to me what the expected behaviour would be if more than one dataset is present. So it will be good to hammer this down.

dbujold: I think the right behaviour would be to show in bento_public one dataset content at a time. (Low priority, but there could be a way to select between available datasets which one is visible in Bento_public. It also makes sense for Provenance metadata, which resolves around one specific dataset)

PR #365, opened 2022-12-08
Add data type labels and delist mcode

davidlougheed:
adds `KATSU_PHENOPACKET_LABEL` env var for configuring phenopacket label


brouillette (approved):
Works on my machine!

PR #369, opened 2022-12-19
Improve error handling when ingestion schema validation fails

davidlougheed:
Here, the issue was that any failing schema checks were discarded instead of throwing an error.
I also decided to move the checks up before trying to create any objects.


gsfk (approved):
Works well, now returns HTTP 400 when ingest fails, and web shows "System Error" in the workflow runs.

Also no sign of partial ingest of partially correct files, for either phenopackets or experiments.

PR #376, opened 2023-01-31
Release 2.18

davidlougheed:
* Use Poetry for dependency management
* Add starts/ends-with/case-insensitive-like search
* Update service-info
* Fix dev-related git issues
* Nicer Docker permissions setup
* Improve cBioPortal workflows and code


gsfk:
Thanks!... `#isw`, `#iew` appear to work as expected (although I had to manually bump the bento_lib version)

This raises an unexpected discussion question (which we don't need to resolve immediately): My first few attempts at using these operators were forbidden!... because I tried it on fields where our phenopackets search spec required katsu "search_optional_eq" operators (rather than search_optional_str, which I was allowed to use). But the motivation for requiring one or the other is not always clear. Is it simply Bento user hand-holding?

davidlougheed:
A mix of hand-holding, whatever makes sense for the field in question, and a primordial mechanism for censored discovery - there are aspects we can/should remove from the search system, and adjustments we can make to the allowed operations in the katsu search schemas if necessary

davidlougheed:
> although I had to manually bump the bento_lib version

@gsfk just FYI this branch does switch Katsu to poetry, which manages dependencies differently vs. pip - this'll be rolling out to all python services eventually since it makes juggling dev/prod dependencies much easier... takes a while to get used to though.

gsfk:
So far I've happily ignored the rest of the pr and just stolen the single commit that I wanted. Currently updating to 2.10 and will drag myself into 2.11 when I get sick of that.

davidlougheed:
poetry is unrelated to 2.10 vs 2.11; some 2.10 services already use it - as a heads up

noctillion (approved):
Looks good and works as expected. I can't see any issues or concerns with the changes made. Great job!

davidlougheed:
thanks for reviewing!

davidlougheed:
note to self: fix readme instructions before merging

v-rocheleau (approved):

PR #377, opened 2023-02-03
Fix categorical counts for public stats endpoint

davidlougheed:
No longer leaks 0-count categories if using derived-from-DB category labels.


gsfk (approved):
Looks good. I don't have any data immediately on hand that was doing this kind of leak, but I can confirm at least that nothing else in the response is broken.

PR #384, opened 2023-03-07
Fix null individual_id in subject-associated biosamples; add logging for individuals public query endpoint

davidlougheed:
* feat: user-specifiable log level
* chore: add logging for individual public query endpoint
* fix: non-present `individual_id` in biosample ingest documents resulting in biosamples not being associated correctly with phenopacket subjects


brouillette (approved):
Tested and works as expected 👍

PR #386, opened 2023-03-22
Update docs setup, remove virtualenv use in dev image, add VSCode debug support

davidlougheed:


docs/requirements.txt
v-rocheleau, line 1: We dont use poetry for the docs?

bento.dev.Dockerfile
```diff
@@ -1,15 +1,23 @@
...
+LABEL org.opencontainers.image.description="Local development image for Katsu."
+LABEL devcontainer.metadata='[{ \
+  "remoteUser": "bento_user", \
+  "customizations": { \
+    "vscode": { \
+      "extensions": ["ms-python.python", "eamodio.gitlens"], \
+      "settings": {"workspaceFolder": "/app"} \
+    } \
+  } \
+}]'
```
v-rocheleau, line 12: Could we have a short section in the README explaining how to use this? Looks neat!

v-rocheleau (commented):
Looks good, would just like to know more about the vscode thingy

docs/requirements.txt
davidlougheed: it's processed by readthedocs, I didnt feel like customizing their build process...

bento.dev.Dockerfile
davidlougheed: will do!

davidlougheed: added

v-rocheleau (approved):

PR #392, opened 2023-04-12
Sonchau/moh part 29

SonQBChau:
What changes:
- fix the bug in the model Donor to allow null value on arrayfield
- change convert script, now it will run both synthetic data and fixtures data and ask for dataset size
- add medium dataset schemas, raw data, synthetic fixture and new relationship JSON

How to test:
- drop the db and create again
- run migrate
- run convert.py for medium set
- load new data into db


SonQBChau:
create by mistake, should PR to candig

PR #393, opened 2023-04-17
Codecov github actions

v-rocheleau:
Codecov's pypi package was [deprecated](https://about.codecov.io/blog/message-regarding-the-pypi-package/#:~:text=On%20April%2012%2C%202023%2C%20Codecov,used%20package%20from%20active%20support.&text=As%20this%20package%20was%20unmaintained,to%20be%20made%20functional%20again) last week.

This PR removes the dependency on the old codecov python package, and replaces it with a Github Action step.


pyproject.toml
```diff
@@ -57,7 +57,6 @@ tabulate = "^0.9.0"
 uritemplate = "^4.1.1"
 
 [tool.poetry.group.dev.dependencies]
-codecov = "^2.1.12"
```
davidlougheed, line 60: this needs a corresponding (should be automated by poetry) removal from the lockfile

davidlougheed (changes requested):

pyproject.toml
v-rocheleau: Good catch, didn't know I had to run `poetry lock` for it to update

davidlougheed (approved):
lgtm

pyproject.toml
davidlougheed: there are commands that remove an entry from pyproject.toml and poetry.lock at once (see poetry docs), but if you remove an item by hand you need the lock command yeah

PR #399, opened 2023-05-01
Features/explorer biosample experiment search

noctillion:
Biosample and experiment data retrieval and export

This pull request introduces the handling of biosample and experiment data associated with phenopackets for displaying in bento-web in explorer tables.

Changes include:

- Modification of get_biosamples_experiments function that retrieves biosample and experiment data associated with a given phenopacket ID. This function helps to simplify the retrieval and display of this information in the application.

- Introduction ExperimentBatchViewSet provides a more flexible way to handle batch operations on experiments. Experiments can be fetched based on a list of IDs through a POST request.

- Implementation of BiosampleBatchViewSet, which allows handling of batch operations on biosamples. Biosamples can be fetched based on a list of IDs through a POST request.

- Addition of two new CSV renderers, BiosamplesCSVRenderer and ExperimentCSVRenderer, which extend the JSONRenderer class. These renderers enable users to export biosample and experiment data in CSV format.


README.md
```diff
@@ -1,4 +1,4 @@
-# Katsu Metadata Service
+# Katsu Metadata ServiceX
```
v-rocheleau: Remove 'X'

chord_metadata_service/chord/views_search.py
```diff
@@ -305,6 +308,20 @@ def mcodepacket_query_results(query, params, options=None):
     return queryset
 
 
+def get_biosamples_experiments(phenopacket_ids):
+    biosamples = Biosample.objects.filter(phenopacket__subject_id__in=phenopacket_ids)\
```
v-rocheleau: Looking at the usage of `get_biosamples_experiments` and the filter lookup, shouldnt the param `phenopacket_ids` be named `subject_ids`?

chord_metadata_service/chord/views_search.py
```diff
@@ -427,9 +457,9 @@ def search(request, internal_data=False):
             "data_type": data_type,
             "matches": list(serializer_class(p).data for p in table_objects)
         } for table_id, table_objects in itertools.groupby(
-            queryset,
-            key=lambda o: str(o.table_id)           # object here
-        )
+            queryset if queryset is not None else [],
+            key=lambda o: str(o.table_id)  # object here
+            )
```
v-rocheleau: de-indent

v-rocheleau (changes requested):
Just a quick look

chord_metadata_service/phenopackets/api_views.py
```diff
@@ -170,6 +173,46 @@ class BiosampleViewSet(ExtendedPhenopacketsModelViewSet):
...
+    content_negotiation_class = FormatInPostContentNegotiation
+
+    def _get_filtered_queryset(self, ids_list=None):
+        queryset = m.Biosample.objects.all() \
+            .prefetch_related(*BIOSAMPLE_PREFETCH) \
+            .select_related(*BIOSAMPLE_SELECT_REL) \
+            .order_by("id")
+
+        if ids_list:
+            queryset = queryset.filter(id__in=ids_list)
```
v-rocheleau, line 197: I think we should apply this filter before `prefetch_related` and `select_related` to limit db operations when ids are passed.

v-rocheleau (changes requested):

chord_metadata_service/phenopackets/api_views.py
noctillion: Absolutely, you're correct. Thank you for pointing this out!

chord_metadata_service/chord/views_search.py
```diff
@@ -305,6 +308,20 @@ def mcodepacket_query_results(query, params, options=None):
     return queryset
 
 
+def get_biosamples_experiments(subject_ids):
+    biosamples = Biosample.objects.filter(phenopacket__subject_id__in=subject_ids)\
+        .values(
+            subject_id=F("phenopacket__subject_id"),
+            biosample_id=F("id"),
+            experiment_id=F("experiment__id"),
```
davidlougheed, line 285: there can be more than one experiment on a biosample, so what happens in that case?

noctillion: the function fetches biosample objects that are associated with the list of subject IDs, and it returns a list of dictionaries the represents a relation between a biosample and the experiment. It includes all associated experiments in the results.

chord_metadata_service/chord/views_search.py
```diff
@@ -318,29 +335,42 @@ def phenopacket_query_results(query, params, options=None):
...
+        )
+
+        # Get the biosamples with experiments data
+        phenopacket_ids = [result['subject_id'] for result in results]
+        biosamples_data = get_biosamples_with_experiment_details(phenopacket_ids)
+
+        # Group the biosamples data by subject_id
+        biosamples_with_experiments = defaultdict(list)
+        for b in biosamples_data:
+            biosamples_with_experiments[b["subject_id"]].append({
```
davidlougheed: i would prefer if this was again not called biosamples, since its actually more of 1 experiment per entry. since experiment records have biosample ID already, you could structure it more like

```
{
    ... all fields from experiment
    "biosample": {
         "id": "BIOSAMPLE_ID",
         "tissue": ...
     }
}
```

then this could be called `experiments_with_biosamples` since biosamples are attached for each experiment.

davidlougheed (changes requested):

chord_metadata_service/chord/views_search.py
davidlougheed: can you add a comment saying that this returns a queryset which will give an entry for each experiment that has an associated biosample or something like that? and maybe rename the function to match what it gets assigned to below

noctillion: Added! Thanks for the suggestion!

v-rocheleau (approved):
lgtm

PR #402, opened 2023-05-15
Beacon search endpoint

gsfk:
Create a new katsu endpoint `/api/beacon_search`, an uncensored equivalent to the bento_public search endpoint `/api/public`. Returns a full list of matching individual ids rather than censored counts, so that  beacon can compute the intersection of results with other searches (in particular for variants). Beacon instead of katsu will apply censorship in this case. 

This allows beacon to use katsu's config for search, which gets us:
1. configurable search, only specific searches are permitted, and
2. fast extra properties seach, currently not available elsewhere in katsu

### DIFF
--- a/chord_metadata_service/patients/api_views.py
+++ b/chord_metadata_service/patients/api_views.py
@@ -258,15 +258,14 @@ class BeaconListIndividuals(APIView):
 
     def get(self, request, *args, **kwargs):
         if not settings.CONFIG_PUBLIC:
-            return Response(settings.NO_PUBLIC_DATA_AVAILABLE)
+            return Response(settings.NO_PUBLIC_DATA_AVAILABLE, status=400)
 
         base_qs = Individual.objects.all()
         try:
             filtered_qs = self.filter_queryset(base_qs)
         except ValidationError as e:
             return Response(errors.bad_request_error(
-                *(e.error_list if hasattr(e, "error_list") else e.error_dict.items()),
-            ))
+                *(e.error_list if hasattr(e, "error_list") else e.error_dict.items())), status=400)
 
         tissues_count, sampled_tissues = biosample_tissue_stats(filtered_qs)
         experiments_count, experiment_types = experiment_type_stats(filtered_qs)
--- a/chord_metadata_service/patients/api_views.py
+++ b/chord_metadata_service/patients/api_views.py
@@ -140,12 +140,13 @@ class IndividualBatchViewSet(BatchViewSet):
 )
 class PublicListIndividuals(APIView):
     """
-    View to return only count of all individuals after filtering.
+    View to return count of all individuals after filtering.
+    Optional parameter "response_format=beacon" adds a list of matching ids to the response,
+    and beacon handles censorship in this case
     """
 
-    def filter_queryset(self, queryset):
+    def filter_queryset(self, queryset, qp):
         # Check query parameters validity
-        qp = self.request.query_params
         if len(qp) > settings.CONFIG_PUBLIC["rules"]["max_query_parameters"]:
             raise ValidationError(f"Wrong number of fields: {len(qp)}")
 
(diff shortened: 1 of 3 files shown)


davidlougheed (approved):
lgtm

chord_metadata_service/patients/api_views.py
```diff
@@ -213,3 +213,72 @@ def get(self, request, *args, **kwargs):
...
+    def filter_queryset(self, queryset):
+        # Check query parameters validity
+        qp = self.request.query_params
+        if len(qp) > settings.CONFIG_PUBLIC["rules"]["max_query_parameters"]:
+            raise ValidationError(f"Wrong number of fields: {len(qp)}")
+
+        search_conf = settings.CONFIG_PUBLIC["search"]
+        field_conf = settings.CONFIG_PUBLIC["fields"]
+        queryable_fields = {
+            f"{f}": field_conf[f] for section in search_conf for f in section["fields"]
```
davidlougheed: i think this is a needless format string - it can just be `f:`

chord_metadata_service/patients/tests/test_api.py
```diff
@@ -650,3 +650,52 @@ def test_public_filtering_age_range_min_and_max_no_config(self):
...
+        response_obj = response.json()
+        self.assertIsInstance(response_obj, dict)
+        self.assertEqual(response_obj, settings.NO_PUBLIC_DATA_AVAILABLE)
+
+    @override_settings(CONFIG_PUBLIC=CONFIG_PUBLIC_TEST)
+    def test_beacon_search_response_invalid_search_key(self):
+        response = self.client.get('/api/beacon_search?birdwatcher=yes')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        response_obj = response.json()
+        self.assertEqual(response_obj["code"], 400)
```
davidlougheed: assert equal to `status.HTTP_400_BAD_REQUEST` instead (for these 3 tests)

chord_metadata_service/patients/api_views.py
```diff
@@ -213,3 +213,72 @@ def get(self, request, *args, **kwargs):
...
+                raise ValidationError(f"Invalid value used in query: {value}")
+
+            # recursion
+            queryset = filter_queryset_field_value(queryset, field_props, value)
+
+        return queryset
+
+    def get(self, request, *args, **kwargs):
+        if not settings.CONFIG_PUBLIC:
+            return Response(settings.NO_PUBLIC_DATA_AVAILABLE)
```
davidlougheed: should this not be a 404 instead of a 200? fine to leave as-is for consistency with bento public responses but generally i think in API design it's good to put as much meaning in the status code as possible

davidlougheed (changes requested):
very minor code style things. other than that it looks good to me

chord_metadata_service/patients/api_views.py
gsfk: Oh noes! You want me to pay attention to the code I'm copypasting? :sob:   

(fixed)

chord_metadata_service/patients/tests/test_api.py
gsfk: These are just copies of the old tests, which, yeah, are a bit weird. Changed the response to 400 and updated the tests.

chord_metadata_service/patients/api_views.py
gsfk: Yeah, this was a straight copypaste of the bento_public code, and it struck me as odd also. I won't break the old code, but I'll change the beacon responses to 400 / bad request. (even for missing config, arguably still 400 rather than 404)

davidlougheed: hmm i think of 400s more meaning something is wrong with the request... if the client request is good, I still think this should be a 404 (or maybe a 500 if we treat it as a configuration error)

gsfk: Ah, okay, I actually re-read the documentation for 404 and this sounds reasonable ("missing resource"). 204 (No content) is also possible but probably a little obscure. I'll change it to 404.

davidlougheed (approved):
lgtm

PR #412, opened 2023-06-12
count_threshold conf to resp in public_overview

noctillion:
This PR is related to Feature #1679, it adds count_threshold information from config to public_overview


gsfk (approved):
Looks good. 

Will probably need more refactoring once we tear out the public backend, but this is great for now.

noctillion:
Sure thing, thanks for the review!

PR #413, opened 2023-06-14
patch: fix unwanted origin schema mutation for required extra_properties

v-rocheleau:
**Issue:** The function schema_utils.patch_project_schemas was causing a mutation on the reference of its base_schema argument. This causes bugs after removing project schemas, as if they were not removed from the DB (they are).
**Fix:** Use deepcopy to remove the reference in the local var and prevent the side effect to occur. 

**Steps to test:**
1. Create a project, dataset, and phenopacket table
2. Create an extra properties schema with **required=true**
3. Run an ingestion (it can fail or pass, doesn't matter)
4. Delete the extra properties schema
5. Run a new ingestion (should pass, was failing)


davidlougheed (approved):
lgtm

PR #414, opened 2023-06-19
Add endpoint for fetch extraPropertiesSchemaTypes

noctillion:
In this PR, is introduced the endpoint, /extra_properties_schema_types. It is aimed at fetching and returning the extra properties schema types.

- A new view, extra_properties_schema_types, was added. This view fetches the schema types and returns them as a dictionary.
- The new view is registered in the URL configuration with the path extra_properties_schema_types.


davidlougheed (approved):
lgtm

PR #433, opened 2023-09-12
feat(patients): add individual phenopackets endpoint w/ attachment option

davidlougheed:
adds an endpoint that looks like `/api/individuals/ind:HG00096/phenopackets` to get all phenopackets for an individual, with an optional `attachment=1` query param to get it with a `Content-Disposition: attachment; ...` header for future downloading use

### DIFF
--- a/chord_metadata_service/patients/api_views.py
+++ b/chord_metadata_service/patients/api_views.py
@@ -101,7 +101,7 @@ class IndividualViewSet(viewsets.ModelViewSet):
 
         return super(IndividualViewSet, self).list(request, *args, **kwargs)
 
-    @action(detail=True, methods=["GET"])
+    @action(detail=True, methods=["GET", "POST"])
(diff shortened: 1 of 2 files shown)


gsfk (approved):
Is this any different from katsu `/api/phenopackets/{id}`? It's hard to tell exactly since our current crop of test data doesn't cover the case of multiple phenopackets per individual. 

This does highlight some weirdness from the current response from `/api/individuals/{id}`, which includes a phenopackets array that you might guess would give the same response as the endpoint in this pr, but you'd be wrong, since it does two odd things: 
- it removes all subject fields (sex, age, etc) and moves them to the top level of the response
- it copies (!) the biosamples field to the top level of the response so that all biosample and experiment data is given twice.

noctillion:
It works as expected locally. lgtm

noctillion (approved):

PR #436, opened 2023-09-18
add missing id field to disease autocomplete

gsfk:
disease autocomplete was returning "label" field twice instead of "id" and "label"


davidlougheed (approved):
lgtm

PR #438, opened 2023-09-22
feat: add dataset resources endpoint

davidlougheed:


gsfk (approved):
I'm guessing the new endpoint is `/api/datasets/{dataset_id}/resources`. This works well, although as usual I have lots of barely relevant comments:

- don't these belong as part of the response from `api/datasets/{dataset_id}`? Or perhaps that bloats the response too much. 
- I peeked at the instance-level resources available both from `/api/resources` (works great) and as part of `/api/metadata` (weirdly broken, I guess a job for a separate pr)

davidlougheed:
since these are 'calculated' from phenopackets on-the-fly, I didn't want to put them in the main response, which just includes `additional_resources` (IDs only though for that field!) - but I agree that it would be good when we do the dataset service to rethink how we do resources/ontology stuff

davidlougheed:
https://206.12.92.46/issues/1925

PR #442, opened 2023-09-29
Add experiment type Viral WGS

noctillion:
This PR include changes in experiment schema to add Viral WGS to experiment_type


davidlougheed (approved):
we will need to also bump the version number in `pyproject.toml` before releasing, but this looks good to me.

PR #449, opened 2023-10-23
chore: remove mcode models and references

v-rocheleau:
Removal of the MCode app needs to be performed in a two steps process.

1. First, we need to remove all references to MCode from the other applications, which entails performing the data migrations in our instances (individual mcode fields) to remove the models without losing data.

2. Finally, when all our instances have been migrated according to the first step, we can completely remove the "mcode" application directory from katsu.

Step **1** requires the mcode migrations (`/chord_metadata_service/mcode/migrations`), which is why we can't remove the "mcode" directory right away in **step 1**.

This PR and its release are the first step in this process.


chord_metadata_service/patients/tests/test_api.py
```diff
@@ -561,13 +562,15 @@ def test_public_filtering_extra_properties_multiple_ranges_1(self):
     @override_settings(CONFIG_PUBLIC=CONFIG_PUBLIC_TEST)
     def test_public_filtering_extra_properties_date_range_1(self):
         # extra_properties date range search (only after or before, single value)
+        # Testing with a date of consent from 1 year ago
+        year_of_consent = datetime.now().year - 1
```
davidlougheed: minor: i think i'd prefer if all dates were fixed rather than being based on now() (so basically changing the other side of this test, the test data rather than the test data itself)

chord_metadata_service/mcode/models.py
```diff
@@ -1,342 +0,0 @@
...
-from django.db import models
-from django.db.models import JSONField
-from django.contrib.postgres.fields import ArrayField
-from chord_metadata_service.restapi.models import IndexableMixin
-from chord_metadata_service.phenopackets.models import Gene
-from chord_metadata_service.patients.models import Individual
-from chord_metadata_service.restapi.description_utils import rec_help
-import chord_metadata_service.mcode.descriptions as d
-from chord_metadata_service.restapi.validators import ontology_validator, ontology_list_validator
-from .validators import (
```
davidlougheed, line 11: can we also remove this validators file that mcode.models was importing from?

davidlougheed: actually, there are many other files (schemas, serializers etc) not removed yet in this PR...

davidlougheed (changes requested):

chord_metadata_service/patients/migrations/0005_remove_individual_comorbid_condition_and_more.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 4.2.5 on 2023-10-23 19:04
```
davidlougheed: these should get squashed into a v5_0_0 migration eventually for the v5 (phenov2) release

chord_metadata_service/mcode/models.py
v-rocheleau: Updated, only left the files/folders required for migrations

davidlougheed (approved):
lgtm!

PR #450, opened 2023-10-26
feat: document experiment results ingestion

noctillion:
This PR adds a workflow for document ingestion (pdf, csv, etc) as part of experiment results.


chord_metadata_service/chord/workflows/metadata.py
```diff
@@ -147,6 +149,30 @@ def json_file_output(id_: str, output_name: Optional[str] = None):
                 }
             ]
         },
+        WORKFLOW_READDOC: {
```
davidlougheed: the `readdoc` name doesn't really make sense – it should be `document` I think

chord_metadata_service/chord/workflows/metadata.py
```diff
@@ -147,6 +149,30 @@ def json_file_output(id_: str, output_name: Optional[str] = None):
...
+            "name": "Document",
+            "description": "XThis workflow will send document files to the specified ingestion service.",
```
davidlougheed: description doesn't really make sense - should be maybe something like

> This workflow ingests into DRS documents which have been already listed as experiment results.`

chord_metadata_service/chord/workflows/wdls/readdoc.wdl
```diff
@@ -0,0 +1,48 @@
+version 1.0
```
davidlougheed: rename everything here from readdoc to a new name like mentioned above

davidlougheed (changes requested):

chord_metadata_service/chord/workflows/metadata.py
noctillion: Perfect!!

chord_metadata_service/chord/workflows/metadata.py
noctillion: Sure, it was tentative. Thank you for the insigths!!

chord_metadata_service/chord/workflows/metadata.py
```diff
@@ -147,6 +149,31 @@ def json_file_output(id_: str, output_name: Optional[str] = None):
...
+                                   ".jpeg", ".jpg", ".png", ".gif", ".md", ".mp3", ".m4a", ".mp4"]
+                },
+                DRS_URL_INPUT,
+            ],
+            "outputs": [
+                {
+                    "id": "document_files_out",
+                    "type": "file",
+                    "value": "{document_files_out}"
+                },
```
davidlougheed: in this case, there shouldn't be any outputs defined here - WES ignores these anyway currently

davidlougheed: (this part of workflow definitions is going away in the future)

chord_metadata_service/chord/workflows/wdls/document.wdl
```diff
@@ -0,0 +1,48 @@
...
+                file_path = file,
+                drs_url = drs_url,
+                project_id = project_id,
+                dataset_id = dataset_id,
+                token = secret__access_token
+        }
+    }
+
+    output {
+        Array[String] document_files_out = post_to_drs.response_message
```
davidlougheed: this is misnamed - it isn't file output, just responses - maybe call this `drs_responses` instead of `document_files_out`

davidlougheed (changes requested):
getting close! a couple last reqs

davidlougheed (approved):
lgtm

PR #457, opened 2023-11-13
chore: implement new workflow system

davidlougheed:
Injected service/config/secret values using new code in Bento Lib / new WES; requires new Authz (0.5.0+); see https://github.com/bento-platform/bento/pull/202


chord_metadata_service/chord/workflows/metadata.py
```diff
@@ -35,234 +28,151 @@
...
+
+# Export workflows -----------------------------------------------------------------------------------------------------
+
+workflow_set.add_workflow(WORKFLOW_CBIOPORTAL, wm.WorkflowDefinition(
+    type="export",
+    name="cBioPortal",
+    description="This workflow creates a bundle for cBioPortal ingestion.",
+    tags=[WORKFLOW_TAG_CBIOPORTAL],
+    file="cbioportal_export.wdl",
+    inputs=[
```
v-rocheleau, line 172: No `ACCESS_TOKEN_INPUT`? Is the cBioPortal export always public?

chord_metadata_service/chord/workflows/wdls/document.wdl
```diff
@@ -30,15 +27,16 @@ task post_to_drs {
...
         String drs_url
-        String project_id
-        String dataset_id
+        String project_dataset
         String token
     }
     command {
+        project_id=$(python3 -c 'print("~{project_dataset}".split(":")[0])')
+        dataset_id=$(python3 -c 'print("~{project_dataset}".split(":")[1])')
         curl -k -X POST \
```
v-rocheleau: I think we should also inject a tag of some sort that indicates if the workflow requests should be secure (PROD) or insecure because of self-signed certs (DEV). Right now all curl commands use the `-k` flag no matter what.

v-rocheleau: Here we could also validate the http code to avoid false workflow completions with 4XX codes

chord_metadata_service/chord/workflows/wdls/readset.wdl
```diff
@@ -1,24 +1,46 @@
...
+        String drs_url
+        String project_dataset
+        String token
+    }
     command {
-        cp '${readset_file_in}' $(basename -- ${readset_file_in})
-        echo $(basename -- ${readset_file_in})
+        project_id=$(python3 -c 'print("~{project_dataset}".split(":")[0])')
+        dataset_id=$(python3 -c 'print("~{project_dataset}".split(":")[1])')
+        curl -k -X POST \
```
v-rocheleau: HTTP code check for workflow failure

v-rocheleau (changes requested):

chord_metadata_service/chord/workflows/metadata.py
davidlougheed: no, just an oversight - fixed

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
```diff
@@ -98,14 +110,17 @@ task get_maf {
...
 
-                # Request from DRS the maf file absolute local path
+                # Request the MAF file from DRS
                 object_id = maf_uri.split("/")[-1].rstrip()
-                drs_object_url = f"~{drs_url}/objects/{object_id}?internal_path=1"
-                response = requests.get(drs_object_url, verify=False)
+                response = requests.get(
+                    f"~{drs_url}/objects/{object_id}?internal_path=1",
+                    headers={"Authorization": "Bearer ~{token}"},
+                    verify=~{true="True" false="False" validate_ssl},
```
v-rocheleau, line 118: String literals are only for cert paths for the `verify` arg

> verify – (optional) Either a boolean, in which case it controls whether we verify the server’s TLS certificate, or a string, in which case it must be a path to a CA bundle to use. Defaults to True.

chord_metadata_service/chord/workflows/wdls/vcf2maf.wdl
```diff
@@ -73,7 +77,7 @@ task katsu_dataset_export_vcf {
         # TODO: handle pagination, i.e. if the `next` property is set, loop
         # over the pages of results
         metadata_url = f"~{katsu_url}/api/experimentresults?datasets={dataset_id}&file_format=vcf&page_size=10000"
-        response = requests.get(metadata_url, verify=False)
+        response = requests.get(metadata_url, verify=~{true="True" false="False" validate_ssl})
```
v-rocheleau, line 80: Same for verify string arg

chord_metadata_service/chord/workflows/wdls/vcf2maf.wdl
```diff
@@ -98,7 +102,7 @@ task katsu_dataset_export_vcf {
                 # Query DRS with the filename to get the absolute file path in
                 # DRS for processing.
                 drs_url = f"${drs_url}/search?name={vcf}&internal_path=1"
-                response = requests.get(drs_url, verify=False)
+                response = requests.get(drs_url, verify=~{true="True" false="False" validate_ssl})
```
v-rocheleau, line 105: Same for verify string arg

chord_metadata_service/chord/workflows/wdls/vcf2maf.wdl
```diff
@@ -212,7 +217,7 @@ task vcf_2_maf {
                 drs_url,
                 headers={"Authorization": "Bearer ~{access_token}"} if "~{access_token}" else {},
                 json=params,
-                verify=False
+                verify=~{true="True" false="False" validate_ssl},
```
v-rocheleau, line 220: Same for verify string arg

chord_metadata_service/chord/workflows/wdls/vcf2maf.wdl
```diff
@@ -320,7 +326,12 @@ task katsu_update_experiment_results_with_maf {
                 "derived_from_data_type": "experiment_result"
             }
         }
-        response = requests.post(metadata_url, headers=headers, json=data, verify=False)
+        response = requests.post(
+            metadata_url,
+            headers=headers,
+            json=data,
+            verify=~{true="True" false="False" validate_ssl},
```
v-rocheleau, line 333: Same for verify string arg

v-rocheleau (changes requested):
Looks good, just not sure about wdl python commands with ssl verify
https://requests.readthedocs.io/en/latest/api/

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
v-rocheleau: It may be fine, from WDL [docs](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#expression-placeholder-options):
> Expression placeholder options are option="value" pairs that precede the expression in an expression command part and **customize the interpolation of the WDL value into the command string being built.**

v-rocheleau (approved):
lgtm!

PR #459, opened 2023-11-21
remove censorship from search code for beacon

gsfk:
Don't apply "max query params" censorship to search code for beacon; beacon handles its own censorship.


davidlougheed (approved):
lgtm; I'm assuming that Beacon itself also checks the threshold, rather than relying on Katsu to do the censorship?

gsfk:
> lgtm; I'm assuming that Beacon itself also checks the threshold, rather than relying on Katsu to do the censorship?

Yes, exactly

PR #462, opened 2023-12-04
fix: prevent unwanted phenopacket dataset update in ingestion

v-rocheleau:
**Issue:**
1. ingest a phenopacket (with id=1) in dataset A
2. create dataset B
3. ingest the same phenopacket (with id=1) in dataset B
4. no duplicate PK error is thrown AND dataset A no longer lists the phenopacket it previously had since it was "stolen" by B.

**Cause:** the `ingest_phenopacket` function was not strict enough in the phenopacket creation and would instead end up updating the phenopacket row to change its dataset FK.

**Solution:** use `get_or_create` instead so that an IntegrityError may be raised if a duplicate ID is used.

### DIFF
--- a/chord_metadata_service/chord/ingest/phenopackets.py
+++ b/chord_metadata_service/chord/ingest/phenopackets.py
@@ -319,7 +319,12 @@ def ingest_phenopacket(phenopacket_data: dict[str, Any],
     if phenopacket_data.get("files", []):
         logger.warning("Found files in phenopacket.files are not ingested by Katsu.")
 
+    # Abort the ingestion if the phenopacket's ID exists in the DB
     phenopacket_id = phenopacket_data.get("id")
+    if pm.Phenopacket.objects.filter(id=phenopacket_id).exists():
+        error_msg = f"Cannot ingest Phenopacket with ID {phenopacket_id}, ID already exists in the database."
+        logger.error(error_msg)
+        raise IngestError(error_msg)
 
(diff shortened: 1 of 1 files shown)


chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -376,19 +375,16 @@ def ingest_phenopacket(phenopacket_data: dict[str, Any],
     # Attach resources to the metadata object
     meta_data_obj.resources.set(resources_db)
 
-    # Create the phenopacket object...
-    new_phenopacket = pm.Phenopacket(
-        id=new_phenopacket_id,
+    # Get the phenopacket if one exists for given ID, or create a new one.
+    new_phenopacket, created = pm.Phenopacket.objects.get_or_create(
```
davidlougheed: not necessarily a `new_phenopacket` anymore (except, it will actually always be new - see the review comment; but then the comment here is wrong since it won't get that phenopacket itll just error) - also `created` is never used

davidlougheed (commented):
looks good overall - the `get_or_create` call is good for throwing an error, but will never actually `get` since the metadata object is always created from scratch - so it'll just throw an error if a phenopacket already exists with the ID.

could just manually raise an ingest error if a phenopacket with the ID already exists, maybe that's a bit clearer?

davidlougheed (approved):
lgtm

PR #463, opened 2023-12-08
chore: rm enum values for experiment_result.genome_assembly_id

davidlougheed:
In order to support the reference service, this should be free-form (de facto a 'foreign key' to the reference service.) In the future, this should be validated on ingestion against the reference service.


v-rocheleau (approved):
Lgtm!

PR #464, opened 2023-12-11
chore: remove hard-coded page caching

davidlougheed:


chord_metadata_service/phenopackets/api_views.py
```diff
@@ -97,9 +89,6 @@ class BiosampleViewSet(ExtendedPhenopacketsModelViewSet):
     post:
     Create a new biosample
     """
-    queryset = m.Biosample.objects.all() \
-        .prefetch_related(*BIOSAMPLE_PREFETCH) \
-        .order_by("id")
```
v-rocheleau, line 102: Why was this removed?

v-rocheleau (commented):
Looks good, one question

chord_metadata_service/phenopackets/api_views.py
davidlougheed: not visible in the diff, but it's a duplicate - if you expand below you can see its there a second time

v-rocheleau (approved):
lgtm!

PR #466, opened 2023-12-22
fix genomic interpretation ingestion

v-rocheleau:
Issue: The Interpretation/Diagnosis relationship (many-to-one) is incorrect. When ingesting Interpretations with Diagnoses that share the same Disease, GenomicInterpretations are associated to the wrong Diagnosis.

Solution: Change the Interpretation/Diagnosis relationship to a one-to-one


chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -236,14 +236,14 @@ def get_or_create_genomic_interpretation(gen_interp: dict) -> pm.GenomicInterpre
...
-    gen_obj, _ = pm.GenomicInterpretation.objects.get_or_create(
+    gen_obj, created = pm.GenomicInterpretation.objects.get_or_create(
         interpretation_status=gen_interp["interpretation_status"],
         gene_descriptor=gene_descriptor,
         variant_interpretation=variant_interpretation,
         extra_properties=gen_interp.get("extra_properties", {}),
     )
 
-    if related_obj:
+    if related_obj and created:
```
davidlougheed: why this change? I'm not sure what the original problem was, but if the object is re-used and the related object doesn't already have the object as an interpretration, shouldn't it still be added?

chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -262,26 +262,36 @@ def get_or_create_disease(disease) -> pm.Disease:
...
-        for gen_interp
-        in genomic_interpretations_data
-    ]
-    # disease = pm.Disease.objects.get_or_create(diagnosis["disease"])
-    diag_obj, _ = pm.Diagnosis.objects.get_or_create(
+def get_or_create_interpretation_diagnosis(interpretation: dict) -> pm.Diagnosis:
+    diagnosis = interpretation.get("diagnosis")
+
+    if not diagnosis:
+        return
```
davidlougheed, line 280: this should use a logger to log a warning/error and in the future generate an ingestion warning in the report, i think

davidlougheed: also this violates the return type hint, since the type hint says there will always be a returned Diagnosis

davidlougheed (changes requested):

chord_metadata_service/chord/ingest/phenopackets.py
v-rocheleau: Removed, I was testing stuff out and forgot to remove this check.

chord_metadata_service/chord/ingest/phenopackets.py
v-rocheleau: An Interpretation may have no Diagnosis, so this shouldn't be logged as an error. The function creates a Diagnosis from an Interpretation if it contains one, otherwise it returns None. I corrected the return type.

davidlougheed (approved):
looks good to me - same comment about maybe prefixing the `related_type`

davidlougheed (commented):
I like the logic for the extra props prefixes! maybe we should factor out `__` and the builder fn for prop names (i.e., `__` + name) into constants/a function but the logic looks solid!

davidlougheed (approved):
lgtm

PR #468, opened 2024-01-17
feat(dataset): dats_file field changed from string to json

v-rocheleau:
- Changes the `Dataset.dats_file` field from string to JSON
- Compatibility: `DatasetViewSet` accepts dataset creation requests with `dats_file` as a JSON or string value, but will now parse string values to JSON before creating the DB row.


chord_metadata_service/chord/migrations/0007_v6_0_1.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.2.7 on 2024-01-16 19:06
```
davidlougheed: technically this is a feature, not a bugfix, so it should be for 6.1.0

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #469, opened 2024-01-17
fix(models): make biosample.sampled_tissue field nullable

v-rocheleau:
Patch for the Phenopacket V2 migration.
The `biosample.sampled_tissue` was required in V1, but is now optional.


davidlougheed (approved):
lgtm

chord_metadata_service/phenopackets/migrations/0009_v6_0_1.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.7 on 2024-01-17 18:06
```
davidlougheed: change to 6.1.0 since the other PR is a feature (allows non-string JSON to be provided for DATS)

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #473, opened 2024-01-25
lint: imports, dependencies, modern type hints, exports

davidlougheed:


pyproject.toml
```diff
@@ -34,7 +34,7 @@ classifiers = [
 python = "^3.10.0"
 bento-lib = {extras = ["django"], version = "^11.3.0"}
 Django = "^4.2.9"
-django-autocomplete-light = "^3.9.4"
+django-autocomplete-light = "^3.11.0rc1"
```
v-rocheleau, line 37: Any reason to use a RC version?

v-rocheleau (commented):
Lgtm, just a question regarding version change

pyproject.toml
davidlougheed: future django 5 support + official Django 4.2 support. they didn't even release a 3.10 non-rc version, so I don't really understand how their versions work - they maybe seem to be misunderstanding semver.

davidlougheed: asked about it here: https://github.com/yourlabs/django-autocomplete-light/issues/1351

v-rocheleau (approved):
LGTM

PR #478, opened 2024-02-01
fix: dataset update did not parse dats strings

v-rocheleau:


davidlougheed (approved):
lgtm. are there tests for the DATS create/endpoint functions for both object-form DATS and stringified DATS?

v-rocheleau:
> lgtm. are there tests for the DATS create/endpoint functions for both object-form DATS and stringified DATS?

The `/api/datasets/<id>/dats` endpoint only accepts GET requests.

davidlougheed:
I meant the main dataset update/create points

v-rocheleau:
Ah yeah I added tests on that one in the last dats PR

PR #483, opened 2024-02-07
chore!: impose more lenient ID regex across routers/schemas

davidlougheed:
This is for models which have view-sets and user-settable IDs in ingestion.


v-rocheleau (approved):
LGTM!

gsfk (approved):
works for me!

PR #490, opened 2024-03-15
fix(workflows): cBioPortal export workflow fixes + updates

davidlougheed:


chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
```diff
@@ -45,15 +41,14 @@ task dataset_id_from_project_dataset {
         python3 -c 'print("~{project_dataset}".split(":")[1], end="")'
     >>>
     output {
-        String dataset_id = stdout()
+        String dataset_id = read_string(stdout())
```
v-rocheleau, line 44: Why is this necessary?

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
```diff
@@ -62,23 +57,17 @@ task katsu_dataset_export {
     # command block (tested with womtool-v78). Using triple angle braces made
     # interpolation more straightforward.
     command <<<
-        # Export results at export_path and returns http code 200 in case of success
-        RESPONSE=$(curl -X POST ~{true="" false="-k" validate_ssl} -s -w "%{http_code}" \
+        # Export results; Katsu returns a .tar.gz file
+        curl -X POST ~{true="" false="-k" validate_ssl} -s --fail-with-body \
```
v-rocheleau, line 61: Nice, I didn't know about the --fail-with-body flag, probably useful for other workflows as well.

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
```diff
@@ -62,23 +57,17 @@ task katsu_dataset_export {
...
             -H "Authorization: Bearer ~{token}" \
-            -d '{"format": "cbioportal", "object_type": "dataset", "object_id": "~{dataset_id}", "output_path": "~{run_dir}"}' \
-            "~{katsu_url}/private/export")
-
-        if [ "${RESPONSE}" != "204" ]
-        then
-            echo "Error: Metadata service replied with HTTP code ${RESPONSE}" 1>&2  # to stderr
-            exit 1
-        fi
-        echo ${RESPONSE}
```
v-rocheleau, line 77: I think the workflow could fail silently now, no?

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
```diff
@@ -87,21 +76,24 @@ task katsu_dataset_export {
...
-        String run_dir
         String dataset_id
         String token
         Boolean validate_ssl
+        File export_data
     }
 
     command <<<
+        # Extract exported data
+        tar -xzvf ~{export_data}
```
v-rocheleau, line 87: Maybe we should check if the extracted data exists?

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
davidlougheed: it didn't work before; `stdout()` resolves to the path of stdout, not the actual stdout content

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
davidlougheed: I think it shouldn't because of the --fail-with-body, although it won't have the `Error: Metadata service replied with HTTP code` print now...

chord_metadata_service/chord/workflows/wdls/cbioportal_export.wdl
davidlougheed: I think having this as a non-optional WDL input should (?) guarantee the path exists at least, but maybe it doesn't. It's difficult to figure out how this actually behaves.

v-rocheleau (approved):
lgtm!

PR #492, opened 2024-03-20
chore: update debugpy JSON for VS code

davidlougheed:


gsfk (approved):
Debugger seems to work fine. 

VS no longer opens the workspace folder properly, but this is separate issue with the Dockerfile.

PR #495, opened 2024-03-25
fix: add missing disease fields

v-rocheleau:


chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -264,8 +264,12 @@ def get_or_create_genomic_interpretation(gen_interp: dict) -> pm.GenomicInterpre
 def get_or_create_disease(disease) -> pm.Disease:
     d_obj, _ = pm.Disease.objects.get_or_create(
         term=disease["term"],
+        excluded=disease.get("excluded"),
```
davidlougheed: per phenopackets v2 spec, if this isn't present it defaults to "false" - so should this be `get("excluded", False)` and changed in the model to be non-null + default False?

chord_metadata_service/phenopackets/migrations/0010_v7_0_3.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 4.2.11 on 2024-03-26 20:10
```
davidlougheed: upcoming katsu release is 8.0.0 assuming the cBioPortal / auth changes make it in for v16.

davidlougheed (changes requested):
migration rename needed; otherwise looks good to me!

davidlougheed (approved):
lgtm

PR #496, opened 2024-03-27
refact!: discovery module + revised overview endpoints

davidlougheed:
BREAKING CHANGE: new formats for various overview endpoints in order to share more code.

beacon: https://github.com/bento-platform/bento_beacon/pull/79

web: https://github.com/bento-platform/bento_web/pull/392

PR #498, opened 2024-04-08
feat: json fields charts and search

v-rocheleau:
Allows the use of JSONField with array values for public charts and search.
### String field props
**mapping** : must end at the JSONField's name
**group_by** (NEW) : mapping-like path pointing to the element key within the JSONField

### Number field props
**mapping** : must end at the JSONField's name
**group_by** (NEW) : mapping-like path pointing to the element key within the JSONField
**group_by_value** (NEW) : value to perform grouping on (e.g. only get measurements where assay/id is equal to "NCIT:C200479")
**value_mapping** (NEW) : mapping-like path pointing to the number key for elements of the JSONField

### Example fields config
```JSON
{
    "fields": {
        "diagnostic_markers": {
            "mapping": "individual/phenopackets/biosamples/diagnostic_markers",
            "group_by": "label",
            "title": "Diagnostic Markers",
            "description": "Markers used for diagnosis",
            "datatype": "string",
            "config": {
                "enum": null
            }
        },
        "measurement_type": {
            "mapping": "individual/phenopackets/measurements",
            "group_by": "assay/label",
            "title": "Measurement types",
            "description": "types of measurements performed",
            "datatype": "string",
            "config": {
                "enum": null
            }
        },
        "measurement_tumor_length": {
            "mapping": "individual/phenopackets/measurements",
            "group_by": "assay/id",
            "group_by_value": "NCIT:C200479",
            "value_mapping": "value/quantity/value",
            "title": "Tumor lengths",
            "description": "measured tumor lengths in millimeters",
            "datatype": "number",
            "config": {
                "minimum": 0,
                "maximum": 200,
                "bin_size": 20,
                "taper_left": 0,
                "taper_right": 200,
                "units": "mm"
            }
        }
    }
}
```


davidlougheed:
this will need to be reconciled with https://github.com/bento-platform/katsu/pull/496 which moves a lot of functions around

chord_metadata_service/discovery/fields.py
```diff
@@ -91,16 +65,19 @@ async def get_distinct_field_values(field_props: DiscoveryFieldProps, low_counts
...
+    values_with_counts = model.objects.values_list(field_query).annotate(count=Count(field))
+
     return [
         val
-        async for val, count in (
-            model.objects
-            .values_list(field)
-            .annotate(count=Count(field))
-        )
+        async for val, count in (values_with_counts)
```
davidlougheed: don't need parentheses

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -1,4 +1,54 @@
-from typing import Iterator
+from typing import Any, Iterator, Type
+from django.db.models import Q, Func, BooleanField, F, Value, Model, JSONField
+
+from chord_metadata_service.discovery.model_lookups import PUBLIC_MODEL_NAMES_TO_MODEL
+
+# from chord_metadata_service.discovery.fields import get_model_and_field
```
davidlougheed: rm?

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -1,4 +1,54 @@
...
+    output_field = JSONField()
+
+
+def get_jsonb_path_query(field: str, json_path: str, is_array=True, is_mapping=True):
+    field_operator = "$[*]" if is_array else "$"
+    query_path = mapping_to_json_path(json_path) if is_mapping else json_path
+    return JSONBPathQuery(F(field), Value(f"{field_operator}.{query_path}"))
+
+
+def get_public_model_name_and_field_path(field_id: str) -> tuple[str, tuple[str, ...]]:
```
davidlougheed, line 26: why were these methods moved here? i assume to solve a circular import of some type

chord_metadata_service/discovery/fields.py
```diff
@@ -338,22 +329,34 @@ def filter_queryset_field_value(qs: QuerySet, field_props, value: str):
...
+    model, field = f_utils.get_model_and_field(
         field_props["mapping_for_search_filter"] if "mapping_for_search_filter" in field_props
         else field_props["mapping"]
     )
 
+    group_by = field_props.get("group_by")
+
     if field_props["datatype"] == "string":
-        condition = {f"{field}__iexact": value}
+        if group_by:
```
davidlougheed, line 340: missing tests for this - is it relatively straightforward, or too complex to bother?

chord_metadata_service/discovery/fields.py
```diff
@@ -338,22 +329,34 @@ def filter_queryset_field_value(qs: QuerySet, field_props, value: str):
...
     elif field_props["datatype"] == "number":
         # values are of the form "[50, 150)", "< 50" or "≥ 800"
 
         if value.startswith("["):
             [start, end] = [int(v) for v in value.lstrip("[").rstrip(")").split(", ")]
-            condition = {
-                f"{field}__gte": start,
-                f"{field}__lt": end
-            }
+            if json_range_condition := f_utils.get_json_range_condition(field_props, start, end):
```
davidlougheed, line 351: same here

davidlougheed (changes requested):

chord_metadata_service/discovery/fields_utils.py
v-rocheleau: exact

chord_metadata_service/discovery/fields.py
```diff
@@ -91,16 +65,19 @@ async def get_distinct_field_values(field_props: DiscoveryFieldProps, low_counts
...
 
-    model, field = get_model_and_field(field_props["mapping"])
+    model, field = f_utils.get_model_and_field(field_props["mapping"])
     threshold = get_threshold(low_counts_censored)
 
+    field_query = field
+    if group_by := field_props.get("group_by"):
+        # JSONField containing an array
+        # use jsonb_path_query field expression
+        field_query = f_utils.get_jsonb_path_query(field, group_by)
```
davidlougheed, line 74: this branch is still untested; if it's annoying to do so then no worries though.

chord_metadata_service/discovery/fields.py
```diff
@@ -189,18 +166,31 @@ async def get_month_date_range(field_props: DiscoveryFieldProps) -> tuple[str |
...
-    whens = [
-        When(
-            **{f"{field}__gte": floor} if floor is not None else {},
-            **{f"{field}__lt": ceil} if ceil is not None else {},
-            then=Value(label),
-        )
-        for floor, ceil, label in f_utils.labelled_range_generator(field_props)
-    ]
+    if group_by and group_by_value and value_mapping:
+        # JSONField array specific
```
davidlougheed: reading this code, it isn't immediately clear why it applies to just JSON fields. I guess it's because `group_by` is only specified as a separate prop in the config to drill into JSON fields (maybe in search v2 this won't exist then, since we can use the JSON schema to detect if we're drilling into a JSON field)? maybe add a comment clarifying

chord_metadata_service/discovery/fields.py
```diff
@@ -2,47 +2,21 @@
...
+from django.db.models import Case, CharField, Count, F, Func, IntegerField, QuerySet, When, Value, Q
 from django.db.models.functions import Cast
-from typing import Any, Mapping, Type
+from typing import Any, Mapping
 
 from ..logger import logger
 
 from . import fields_utils as f_utils
 from .censorship import get_threshold, thresholded_count
 from .fields_utils import monthly_generator
```
davidlougheed: I know it's unrelated to this PR somewhat, but can you replace the `monthly_generator` use with an `f_utils.monthly_generator` and remove this line? it's bothering me looking at the diff below... 

otherwise, lgtm!

chord_metadata_service/discovery/fields.py
```diff
@@ -338,38 +330,55 @@ def filter_queryset_field_value(qs: QuerySet, field_props, value: str):
...
+            else:
+                condition = Q(**{
+                    f"{field}__gte": start,
+                    f"{field}__lt": end
+                })
         else:
             [sym, val] = value.split(" ")
             if sym == "≥":
-                condition = {f"{field}__gte": int(val)}
+                if json_range_condition := f_utils.get_json_range_condition(field_props, min=int(val)):
```
davidlougheed, line 362: just checking, this is false-y when `Q()` is empty?

v-rocheleau: Yes, the `__bool__` function of `Q` is `len(children)`, so it is false when it is empty

chord_metadata_service/restapi/utils.py
```diff
@@ -3,6 +3,7 @@
 
 
 COMPUTED_PROPERTY_PREFIX = "__"
+MAPPING_SEPARATOR = "/"
```
davidlougheed: duplicate declarations; consolidate to one location.

davidlougheed (changes requested):

chord_metadata_service/restapi/utils.py
v-rocheleau: was not even used :sweat_smile:

davidlougheed:
@v-rocheleau `COMPUTED_PROPERTY_PREFIX` is still present twice - can you consolidate this? (it might be from before)

davidlougheed (approved):
lgtm

PR #508, opened 2024-04-30
feat: project and dataset scoped discovery config

v-rocheleau:
### Discovery configuration hierarchy
Discovery configs can be provided at the following levels:
- Node (classic):
   - Mounted JSON file
   - None => no public data
- Project:
   - Optional at project creation
   - None => node discovery fallback
- Dataset:
   - Optional at dataset creation
   - None => project discovery fallback

When querying a scoped Katsu endpoint, the user can scope the query for a specific project/dataset by using the `project` and/or `dataset` query parameters. This does not apply to endpoints that expect URL parameters for project/dataset IDs.

Katsu handles scoped queries this way:
1. **`{endpoint}?dataset={dataset_uuid}`**
    - uses the dataset discovery config
2.  **`{endpoint}?project={project_uuid}&dataset={dataset_uuid}`**
    - uses the dataset discovery config
    - the dataset MUST belong to the project
3. **`{endpoint}?project={project_uuid}`**
    - uses the project's discovery config
5. **`{endpoint}` **OR** `{endpoint}?project=""&dataset=""`**
    - uses the node's discovery config
    - compatibility with older _bento_web_ and _bento_public_ versions

### Katsu endpoints scoping
The following endpoints can now be queried with the `project` and/or `dataset` query parameters.
1. `public_search_fields`
2. `public_overview`
3. `public_rules`
4. `public`
5. `beacon_search`

### New public endpoint on katsu:

- `public_rules`: returns the scoped public rules (max query params and count threshold)
    - related gateway [PR](https://github.com/bento-platform/bento_gateway/pull/17)
- `schemas/<name>/<subschema>`: returns a JSON-schema or subschema for a given schema name
   - e.g. a client can now get the Phenopacket biosample subschema with this url `api/schemas/phenopacket/biosample`


chord_metadata_service/chord/migrations/0008_dataset_discovery_project_discovery.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 4.2.11 on 2024-04-24 14:10
```
davidlougheed: rename migration for version

chord_metadata_service/discovery/schemas.py
```diff
@@ -0,0 +1,110 @@
+from pathlib import Path
+from chord_metadata_service.restapi.schema_utils import (
+    get_schema_app_id, sub_schema_uri, array_of, base_type, SchemaTypes, enum_of
+)
+
+base_uri = get_schema_app_id(Path(__file__).parent.name)
+
+DISCOVERY_FIELD_SCHEMA = {
+    "$id": sub_schema_uri(base_uri, "discovery_field"),
```
davidlougheed: it would be good for this ID to be a true URL of the schema; but maybe this isn't so straightforward to do right now?

chord_metadata_service/discovery/stats.py
```diff
@@ -2,9 +2,11 @@
 
 from typing import Mapping, Type
 
+from chord_metadata_service.discovery.model_lookups import PUBLIC_MODEL_NAMES_TO_SCOPE_FILTERS
```
davidlougheed: `from .model_lookups` and group with other local relative imports in alphabetical order

chord_metadata_service/discovery/stats.py
```diff
@@ -71,19 +73,35 @@ async def bento_public_format_count_and_stats_list(
     return thresholded_count(total, low_counts_censored), stats_list
 
 
+def get_scoped_queryset(model: Type[Model], project_id: str, dataset_id: str):
```
davidlougheed: add to `__all__`

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -48,6 +48,13 @@ def get_model_and_field(field_id: str) -> tuple[Type[Model], str]:
     return model, field_name
 
 
+def get_public_model_name(model: Type[Model]):
```
davidlougheed: add return type

chord_metadata_service/discovery/model_lookups.py
```diff
@@ -19,3 +19,18 @@
     "biosample": DATA_TYPE_PHENOPACKET,
     "experiment": DATA_TYPE_EXPERIMENT,
 }
+
+PUBLIC_MODEL_NAMES_TO_SCOPE_FILTERS = {
+    "individual": {
+        "project": "phenopackets__dataset__project__identifier",
```
davidlougheed: looking at these deeply nested joins with the django ORM, it seems like it could introduce n+1 query issues... have you looked at if this is happening?

chord_metadata_service/discovery/api_views.py
```diff
@@ -139,11 +181,11 @@ async def public_overview(_request: DrfRequest):
     async def _get_field_response(field_id: str, field_props: dict) -> dict:
         stats: list[BinWithValue] | None
         if field_props["datatype"] == "string":
-            stats = await get_categorical_stats(field_props, low_counts_censored=True)
+            stats = await get_categorical_stats(field_props, project_id, dataset_id, low_counts_censored=True)
```
davidlougheed: this function is typed as `project_id: str` but here `project_id: str | None`.
maybe make the categorical stats function take `str | None` for these and default them to `None` instead, and normalize project/dataset ID getters above with something like `dataset_id = request.query_params.get("dataset") or None`

davidlougheed (changes requested):

chord_metadata_service/discovery/schemas.py
v-rocheleau: Schemas for discovery, phenopacket and experiment now use a true URL as the $id, with the accompanying views to resolve schemas/subschemas

chord_metadata_service/discovery/model_lookups.py
v-rocheleau: Added some "prefetch_related" to avoid n+1

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -48,6 +48,13 @@ def get_model_and_field(field_id: str) -> tuple[Type[Model], str]:
     return model, field_name
 
 
+def get_public_model_name(model: Type[Model]) -> PublicModelNames:
```
davidlougheed, line 51: can do a quick unit test for this including testing the value error

chord_metadata_service/discovery/model_lookups.py
```diff
@@ -1,21 +1,72 @@
...
     "individual": DATA_TYPE_PHENOPACKET,
     "biosample": DATA_TYPE_PHENOPACKET,
     "experiment": DATA_TYPE_EXPERIMENT,
 }
+
+
+class ScopeFilter(TypedDict):
+    filter: str
+    prefetch_related: tuple[str]
+    select_related: tuple[str]
```
davidlougheed: should these be `tuple[str]` or `tuple[str, ...]` for future use?

chord_metadata_service/discovery/schemas.py
```diff
@@ -0,0 +1,108 @@
...
+                "count_threshold": base_type(SchemaTypes.INTEGER),
+                "max_query_parameters": base_type(SchemaTypes.INTEGER)
+            }
+        }
+    },
+    "additionalProperties": False
+}
+
+
+# DATS
```
davidlougheed: missing?

chord_metadata_service/experiments/api_views.py
```diff
@@ -147,3 +147,14 @@ def get_experiment_schema(_request):
     Experiment schema
     """
     return Response(EXPERIMENT_SCHEMA)
+
+
+@api_view(["GET"])
+@permission_classes([AllowAny])
+def get_experiment_subschema(_request, subschema: str):
```
davidlougheed, line 155: add test

chord_metadata_service/experiments/schemas.py
```diff
@@ -198,3 +199,8 @@
     },
     "required": ["id", "experiment_type"]
 }, EXPERIMENT)
+
+# URI referencing
+experiment_resource = Resource.from_contents(EXPERIMENT_SCHEMA)
```
davidlougheed, line 207: these constants are missing from `__all__`

chord_metadata_service/metadata/settings.py
```diff
@@ -50,6 +50,10 @@
 BENTO_CONTAINER_LOCAL = os.environ.get("BENTO_CONTAINER_LOCAL", "false").lower() == "true"
 
 CHORD_URL = os.environ.get("CHORD_URL")  # Leave None if not specified, for running in other contexts
+if CHORD_URL:
+    SCHEMAS_BASE_URL = CHORD_URL + "/api/metadata/api/schemas"
+else:
+    SCHEMAS_BASE_URL = "/chord_metadata_service/schemas"
```
davidlougheed, line 58: this schema URI setup is much better than before, but this hard-codes the Bento/other-deployment path scheme in a way that's a bit inflexible. should we use a `SERVICE_URL_BASE_PATH` environment variable for the `{url...}/api/metadata` portion, like other services? (would require a bento repo change too)

chord_metadata_service/experiments/api_views.py
davidlougheed: also, do the subschemas also have resolvable URIs, or how does that work?

davidlougheed (changes requested):

chord_metadata_service/discovery/schemas.py
v-rocheleau: Removed the comment. I was planning on reworking the way we import/use the DATS schema, but I think this is better left for another PR.

chord_metadata_service/experiments/api_views.py
v-rocheleau: Done. All the subschemas of `Experiment` and `Phenopacket` now have hierarchical and resolvable URIs.

chord_metadata_service/metadata/settings.py
v-rocheleau: Added in https://github.com/bento-platform/bento/pull/247/files

chord_metadata_service/chord/views_search.py
```diff
@@ -553,7 +561,7 @@ async def dataset_summary(request: DrfRequest, dataset_id: str):
     dataset = await Dataset.objects.aget(identifier=dataset_id)
 
     summaries = await asyncio.gather(
-        *map(lambda dt: DATASET_DATA_TYPE_SUMMARY_FUNCTIONS[dt](request, dataset),
+        *map(lambda dt: DATASET_DATA_TYPE_SUMMARY_FUNCTIONS[dt](dataset),
```
davidlougheed: these unused arguments were there to be eventually used for authz, could you put them back for now?

chord_metadata_service/discovery/model_lookups.py
```diff
@@ -1,21 +1,72 @@
...
+class ProjectDatasetScopeFilters(TypedDict):
+    project: ScopeFilter
+    dataset: ScopeFilter
+
+
+PUBLIC_MODEL_NAMES_TO_SCOPE_FILTERS: dict[PublicModelNames, ProjectDatasetScopeFilters] = {
+    "individual": {
+        "project": {
+            "filter": "phenopackets__dataset__project__identifier",
+            "prefetch_related": ("phenopackets__dataset__project")
```
davidlougheed: are these parens supposed to be strings or single item tuples? right now they're just strings wrapped in redundant parens, but I think you want them to be 1-tuples?

chord_metadata_service/restapi/api_views.py
```diff
@@ -59,18 +62,22 @@ async def build_overview_response(phenopackets: QuerySet, experiments: QuerySet)
...
     """
     get:
     Overview of all Phenopackets and experiments in the database - private endpoint
     """
 
     # TODO: permissions based on project - this endpoint should be scrapped / completely rethought
+    try:
+        discovery = await get_request_discovery(request)
+    except DiscoveryConfigException as e:
+        return Response(e.message, status=status.HTTP_404_NOT_FOUND)
```
davidlougheed: I think at least one of this or the patients try/except that's tested above should be tested if it's not too ugly to do so

davidlougheed (changes requested):

chord_metadata_service/discovery/model_lookups.py
v-rocheleau: Added a trailling `,` to make 1-tuples

chord_metadata_service/restapi/api_views.py
v-rocheleau: Added a test.
Will we replace usage of this endpoint with the "public overview" when we bring in the charts from public eventually? Because right now the discovery is not really doing anything on the "private overview".

davidlougheed: yes, the idea is to eventually have a single "overview" endpoint (and for a lot of these endpoints to not care about whether they're "public" or "private" *per se*, more just the level of permissions)

v-rocheleau: Wondering if I should remove discovery scoping from the private overview endpoint and just use the node's config

davidlougheed: you could make it node-only yeah, to keep it exactly backwards-compatible with older versions

davidlougheed: maybe same with search overview...

chord_metadata_service/discovery/utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+async def _get_dataset_discovery(dataset_id: str) -> dict:
+    dataset = await cm.Dataset.objects.aget(identifier=dataset_id)
+    if not dataset.discovery:
+        project = await cm.Project.objects.aget(datasets=dataset_id)
+        return await _get_project_discovery(project=project)
+    return dataset.discovery
+
+
+async def get_discovery(project_id: str = None, dataset_id: str = None) -> DiscoveryConfig:
```
davidlougheed: bad type hints: should be `str | None`

chord_metadata_service/discovery/utils.py
```diff
@@ -0,0 +1,58 @@
...
+from chord_metadata_service.discovery.types import DiscoveryConfig
+from ..chord import models as cm
+
+__all__ = [
+    "get_discovery",
+    "get_request_discovery",
+]
+
+
+async def _get_project_discovery(project_id: str = None, project: cm.Project = None) -> dict:
```
davidlougheed: bad type hints: should be `str | None` and `cm.Project | None`

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #523, opened 2024-07-17
fix: biosamples search total count

v-rocheleau:
Fixes the discrepancy in biosample counts between public-search and beacon.
- compute total count with all categories, even those with missing values or below threshold
- return `stats_list` containing only the categories that respect the threshold.

### DIFF
--- a/chord_metadata_service/discovery/stats.py
+++ b/chord_metadata_service/discovery/stats.py
@@ -60,18 +60,15 @@ async def bento_public_format_count_and_stats_list(
 
     async for q in annotated_queryset:
         label = q["label"]
-        value = thresholded_count(int(q["value"]), discovery, low_counts_censored)
+        raw_value = int(q["value"])
+        thresholded_value = thresholded_count(int(q["value"]), discovery, low_counts_censored)
+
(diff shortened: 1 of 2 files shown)


chord_metadata_service/discovery/stats.py
```diff
@@ -60,18 +60,15 @@ async def bento_public_format_count_and_stats_list(
 
     async for q in annotated_queryset:
         label = q["label"]
-        value = thresholded_count(int(q["value"]), discovery, low_counts_censored)
+        raw_value = int(q["value"])
+        thresholded_value = thresholded_count(int(q["value"]), discovery, low_counts_censored)
```
davidlougheed: reuse `raw_value` as the first param to avoid doing the index-access + int-cast twice

davidlougheed (changes requested):
* bump version to `8.0.1` so we can release right away
* eventually a test might be nice here, but for now we should focus on getting v16 out

davidlougheed (approved):
lgtm - this does make some small cells a bit more obvious; another argument for future quantization maybe.

PR #525, opened 2024-07-18
chore!: update to Django 5 and DRF 3.15

davidlougheed:
Note that this changed the behaviour of trying to create a duplicate project JSON schema (from 500 to 400). Is this acceptable?

### DIFF
--- a/chord_metadata_service/chord/tests/test_api.py
+++ b/chord_metadata_service/chord/tests/test_api.py
@@ -1,5 +1,4 @@
 import json
-from django.db.utils import IntegrityError
 
 from django.urls import reverse
 from rest_framework import status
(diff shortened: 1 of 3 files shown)


v-rocheleau (approved):
Lgtm, the error code change should not impact bento_web, it will handle 400 or 500 as an error.
I did find out that there is a regression for project json-schema creation in bento_web, looking to fix it (not related to this PR).

davidlougheed:
image

getting a strange error here (probably a breaking change to the remote user backend class with django), will investigate later

v-rocheleau (changes requested):
Did my first test wrong, seems to break a number of requests. Investigating.

davidlougheed:
@v-rocheleau pushed a fix to bento lib which makes katsu work (I think)

gsfk (approved):
Hammered as much as I could and didn't find anything specific to this pr. (For example, adding extra property schemas seems to be broken, but I assume this is the regression Victor is talking about above. The same bug is in the current release.)

PR #529, opened 2024-08-07
feat!: authorization support for some (mostly "Bento Public") endpoints

davidlougheed:


davidlougheed:
@v-rocheleau still work to be done, but requesting an initial look at this code

chord_metadata_service/authz/middleware.py
```diff
@@ -0,0 +1,38 @@
...
+pattern_get = re.compile(r"^GET$")
+
+include_pattern_public = (
+    re.compile(r"^(GET|POST|PUT|DELETE)$"),
+    re.compile(r"^/api/(projects|datasets|public|public_overview|public_search_fields|public_dataset|public_rules)$"),
+)
+include_pattern_workflows = (pattern_get, re.compile(r"^(/workflows$|/workflows/)"))
+include_pattern_si = (pattern_get, re.compile(r"^/service-info"))
+include_pattern_schemas = (pattern_get, re.compile(r"^/schemas/.+$"))
+include_pattern_schema_types = (pattern_get, re.compile(r"^/extra_properties_schema_types$"))
```
v-rocheleau, line 25: If I understand correctly, requests matching these patterns will have to pass an authz check.
For now are we assuming the rest of the endpoints are protected by the gateway?

chord_metadata_service/authz/permissions.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class BentoDeferToHandler(BasePermission):
+    def has_permission(self, _request, _view):
+        return True  # we return true, like AllowAny, but we don't mark authz as done - so we defer it to the handler
+
+
+class ReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        return request.method in SAFE_METHODS
```
v-rocheleau, line 41: would be good to cover in tests

chord_metadata_service/authz/types.py
```diff
@@ -0,0 +1,21 @@
...
+
+__all__ = [
+    "Bools",
+    "DataPermissionsDict",
+    "DataTypeDiscoveryPermissions",
+    "FieldDiscoveryPermissions",
+]
+
+
+Bools = tuple[bool, ...]
```
v-rocheleau: unused type, is this for later?

chord_metadata_service/chord/api_views.py
```diff
@@ -31,19 +43,32 @@
...
     pagination_class = LargeResultsSetPagination
     permission_classes = [OverrideOrSuperUserOnly]  # Explicit
 
+    async def get_obj_async(self):
+        return await sync_to_async(self.get_object)()
+
 
 class CHORDPublicModelViewSet(CHORDModelViewSet):
-    permission_classes = [OverrideOrSuperUserOnly | ReadOnly]
+    permission_classes = [BentoAllowAnyReadOnly | BentoDeferToHandler]
```
v-rocheleau: I think this overrides the permission_classes `OverrideOrSuperUserOnly` from `CHORDModelViewSet`, is this intended? If so, we can probably remove it from `CHORDModelViewSet` since it is not being used anywhere else.

chord_metadata_service/chord/api_views.py
```diff
@@ -120,19 +214,53 @@ def _parse_dats(self, request) -> str | None:
...
+            await authz.async_evaluate_one(
+                request,
+                build_resource(project=str(dataset.project_id), dataset=str(dataset.identifier)),
+                P_EDIT_DATASET,
+            )
+        ):
+            return forbidden(request)  # side effect: sets authz done flag
+
+        # Do not allow datasets to change project
+        # TODO
```
v-rocheleau: Good point. We can compare the `request.data['project']` value with `dataset.project_id` and reject if we detect a mismatch.

chord_metadata_service/chord/api_views.py
```diff
@@ -120,19 +214,53 @@ def _parse_dats(self, request) -> str | None:
...
+            return not_found(request)  # side effect: sets authz done flag
+
+        if not (
+            await authz.async_evaluate_one(
+                request,
+                build_resource(project=str(dataset.project_id), dataset=str(dataset.identifier)),
+                P_EDIT_DATASET,
+            )
+        ):
+            return forbidden(request)  # side effect: sets authz done flag
```
v-rocheleau, line 251: Testing looks complex with the authz service mocks, but I think we should at least have coverage on all lines that return a forbidden response.

chord_metadata_service/authz/tests/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+        return self.dt_get("bool", u, *args, **kwargs)
+
+    def dt_authz_counts_get(self, u: str, *args, **kwargs):
+        return self.dt_get("counts", u, *args, **kwargs)
+
+    def dt_authz_full_get(self, u: str, *args, **kwargs):
+        return self.dt_get("full", u, *args, **kwargs)
+
+    def dt_authz_full_post(self, u: str, *args, **kwargs):
+        return self.dt_post("full", u, *args, **kwargs)
```
v-rocheleau: Would give a more expressive name to the `u` argument in these functions

chord_metadata_service/chord/tests/test_api.py
v-rocheleau, line 1: Is there a reason why these tests don't use the `AuthzAPITestCase`? Could factor out some of the mock requests code here.

chord_metadata_service/chord/tests/test_api_data_types.py
```diff
@@ -1,18 +1,77 @@
...
+from .constants import VALID_DATA_USE_1
 
 POST_GET = ("POST", "GET")
 
 DATA_TYPE_NOT_REAL = "not_a_real_data_type"
 
 
-class DataTypeTest(APITestCase):
+class DataTypeHelperTest(TestCase, PermissionsTestCaseMixin):
+    def setUp(self):
```
v-rocheleau: Can extend `PhenoTestCase` class here to factor out this setup code

v-rocheleau (commented):
First pass, looking good so far!

chord_metadata_service/authz/middleware.py
davidlougheed: yes, exactly

chord_metadata_service/authz/tests/helpers.py
davidlougheed: done

chord_metadata_service/chord/tests/test_api_data_types.py
davidlougheed: done

chord_metadata_service/authz/types.py
davidlougheed: removed

chord_metadata_service/authz/middleware.py
davidlougheed: added a comment about this

chord_metadata_service/chord/tests/test_api.py
davidlougheed: done

chord_metadata_service/chord/api_views.py
davidlougheed: dataset updates are now mostly tested - just missing some DATS update test branch cov, which is out of scope

chord_metadata_service/authz/permissions.py
davidlougheed: this pre-dates this PR; it was moved from the `chord` module

chord_metadata_service/chord/api_views.py
davidlougheed: implemented

chord_metadata_service/chord/api_views.py
davidlougheed: removed `CHORDModelViewSet` completely

chord_metadata_service/authz/tests/helpers.py
v-rocheleau, line 1: When in a Katsu dev container, running `./manage.py test` causes connection refused errors:
Connection refused: POST https://bentov2.local/api/authorization/policy/evaluate

chord_metadata_service/cleanup/tests/test_cleanup.py
```diff
@@ -124,7 +126,10 @@ async def test_dataset_delete_api(self):
         self.assertEqual(await clean_individuals(), 0)
         self.assertEqual(await clean_resources(), 0)
 
-        r = await self.async_client.delete(f"/api/datasets/{self.dataset.identifier}")
+        with aioresponses() as m:
+            mock_authz_eval_one_result(m, True)
+            r = await self.async_client.delete(f"/api/datasets/{self.dataset.identifier}")
```
v-rocheleau:
```python
r = self.one_authz_delete(f"/api/datasets/{self.dataset.identifier}")
```

chord_metadata_service/cleanup/tests/test_cleanup.py
```diff
@@ -233,7 +238,10 @@ async def test_experiment_deletion(self):
         self.assertEqual(await ec.clean_experiment_results(), 0)
         self.assertEqual(await ec.clean_instruments(), 0)
 
-        r = await self.async_client.delete(f'/api/datasets/{self.dataset.identifier}')
+        with aioresponses() as m:
+            mock_authz_eval_one_result(m, True)
+            r = await self.async_client.delete(f'/api/datasets/{self.dataset.identifier}')
+
```
v-rocheleau:
```python
r = self.one_authz_delete(f"/api/datasets/{self.dataset.identifier}")
```

chord_metadata_service/discovery/api_views.py
```diff
@@ -93,7 +115,7 @@ async def _counts_for_model_name(mn: str) -> tuple[str, int]:
     }
 )
 @api_view(["GET"])  # Don't use BentoAllowAny, we want to be more careful of cases here.
```
v-rocheleau: outdated comment

chord_metadata_service/discovery/api_views.py
```diff
@@ -104,49 +126,61 @@ async def public_overview(request: DrfRequest):
...
     for public_model_name in counts:
-        if 0 < counts[public_model_name] <= count_threshold:
-            logger.info(f"Public overview: {public_model_name} count is below count threshold")
+        dt = PUBLIC_MODEL_NAMES_TO_DATA_TYPE[public_model_name]
+        rules = get_rules(discovery, dt_permissions[dt])
+        count_threshold = rules["count_threshold"]
+
+        # Extra check for threshold being above 0 to not log warnings for true-0 counts with query:data
+        if 0 < counts[public_model_name] <= count_threshold and count_threshold > 0:
+            logger.info(f"Public overview: {public_model_name} count is below count threshold of {count_threshold}")
```
v-rocheleau: logging the project-dataset scope could be useful for debugging censored counts

chord_metadata_service/discovery/tests/test_api.py
```diff
@@ -108,90 +108,60 @@ def assert_response_section_fields(self, response_obj: dict, config: dict):
...
-
-        # SCOPE: project_a + dataset_a (valid)
-        response_pd_valid = self.client.get(
-            f"{search_fields_url}?project={str(self.id_proj_a)}&dataset={self.id_ds_a}",
-            content_type="application/json",
-        )
-        self.assertEqual(response_pd_valid.status_code, status.HTTP_200_OK)
-        # same as dataset_a
-        self.assertEqual(response_pd_valid.json(), response_d_a.json())
+        subtest_params: list[tuple[DTAccessLevel, str, int, DiscoveryConfig | None]] = [
```
v-rocheleau: nice

chord_metadata_service/discovery/utils.py
v-rocheleau, line 1: I would add some top-level function comments here. These might not be self evident for people who have not touched discovery.

chord_metadata_service/metadata/settings.py
```diff
@@ -51,6 +51,15 @@
...
 
+# SECURITY WARNING: Don't run with AUTHZ_ENABLED turned off in production,
+# unless an alternative permissions system is in place.
+#  - This needs to be here to avoid a circular import with settings.py
+BENTO_AUTHZ_ENABLED: bool = os.environ.get("BENTO_AUTHZ_ENABLED", "true").strip().lower() == "true"
+
+BENTO_AUTHZ_SERVICE_URL: str = (
+    os.environ.get("BENTO_AUTHZ_SERVICE_URL", "http://authz.local").strip().rstrip("/") if BENTO_AUTHZ_ENABLED else ""
+)
+
```
v-rocheleau: Regarding the test errors when in dev mode.
I did a quick test and we can alter `BENTO_AUTHZ_SERVICE_URL` based on `sys.argv` to detect if we are running tests. The following code block allowed me to run all tests in DEV with no errors:
```python
if len(sys.argv) > 1 and sys.argv[1] == 'test':
    BENTO_AUTHZ_SERVICE_URL = "http://authz.local"
```

chord_metadata_service/patients/api_views.py
```diff
@@ -149,34 +159,46 @@ class IndividualBatchViewSet(BatchViewSet):
...
-    # TODO: allow exceeding max query parameters for authorized requests
-    if len(qp) > get_max_query_parameters(discovery, low_counts_censored):
+async def public_discovery_filter_queryset(
+    discovery: DiscoveryConfig, request: DrfRequest, dt_permissions: DataTypeDiscoveryPermissions, queryset: QuerySet
+):
+    # Process query parameters and check validity
+
+    qp: QueryDict = deepcopy(request.query_params)
+
+    # - remove project/dataset (i.e., scope) query parameters
```
v-rocheleau: ah was this messing up the fields count?

chord_metadata_service/patients/api_views.py
```diff
@@ -149,34 +159,46 @@ class IndividualBatchViewSet(BatchViewSet):
...
+        del qp["project"]
+    if "dataset" in qp:
+        del qp["dataset"]
+
+    queryable_fields = get_discovery_queryable_fields(discovery)
+    queried_fields = list(set(qp.keys()))
+
+    overall_permissions, qf_permissions = get_discovery_field_set_permissions(discovery, queried_fields, dt_permissions)
+
+    if len(qp) > get_max_query_parameters(discovery, overall_permissions):
```
v-rocheleau: Is the `set` on `queried_fields` intended to remove duplicates in the query parameters?
If so, we may want to check the number of fields with `len(queried_fields)`

chord_metadata_service/patients/tests/test_api.py
```diff
@@ -415,34 +437,34 @@ def test_public_filtering_extra_properties_1(self):
...
 
     @override_settings(CONFIG_PUBLIC=DISCOVERY_CONFIG_EXTRA_PROPERTIES)
     def test_public_filtering_extra_properties_2(self):
         # extra_properties string search (multiple values)
-        response = self.client.get(
+        response = self.dt_authz_counts_get(
             '/api/public?smoking=Non-smoker&death_dc=deceased&covidstatus=positive'
         )
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
v-rocheleau, line 511: why is this 400 now?

chord_metadata_service/restapi/urls.py
```diff
@@ -91,7 +91,4 @@
     path('public_overview', public_overview, name='public-overview',),
     path('public_dataset', public_dataset, name='public-dataset'),
     path('public_rules', public_rules, name='public-rules'),
-
-    # uncensored endpoint for beacon search using fields from config.json
-    path('beacon_search', individual_views.BeaconListIndividuals.as_view(), name='beacon-search'),
```
v-rocheleau, line 96: is this because Beacon will now use the regular search endpoint?

v-rocheleau (changes requested):

chord_metadata_service/restapi/urls.py
davidlougheed: that's the idea - I'll have to implement a system for it to get its own token though.

chord_metadata_service/patients/api_views.py
davidlougheed: yes, exactly - i'll expand on that a bit in the comment

chord_metadata_service/cleanup/tests/test_cleanup.py
davidlougheed: it's async, which is why it didn't get factored out, but I can create an async version of it.

chord_metadata_service/cleanup/tests/test_cleanup.py
davidlougheed: async here too - needs a diff fn.

chord_metadata_service/patients/api_views.py
davidlougheed: since multiple queries on the same field still (maybe) should count against the # of allowed queries, it may make sense to leave as-is? i'll add a note at least

chord_metadata_service/discovery/api_views.py
davidlougheed: I agree, although technically I think this becomes a log injection vulnerability in the current way it's written if we put the params in with no additional validation

v-rocheleau: Good point. We can add a comment and get back to it when we implement structured logs with user-input validation.

davidlougheed: working on a sub-PR to better structure all of this scope information and do some sanitization.

chord_metadata_service/patients/tests/test_api.py
davidlougheed: it was already returning a 400-code error in the JSON body, but with a 200 HTTP response code - this seems confusing, so I made it consistently an error

chord_metadata_service/metadata/settings.py
davidlougheed: done

chord_metadata_service/patients/api_views.py
```diff
@@ -170,32 +171,51 @@ def get_queryset(self):
 
 
 async def public_discovery_filter_queryset(
-    discovery: DiscoveryConfig, request: DrfRequest, dt_permissions: DataTypeDiscoveryPermissions, queryset: QuerySet
+    discovery: DiscoveryConfig,
+    request: DrfRequest,
+    dt_permissions: DataTypeDiscoveryPermissions,
+    queryset: QuerySet,
+    scope_repr: str,  # We pass disc
```
v-rocheleau: `discovery` and `scope_repr` arguments are derived from the same `ValidatedDiscoveryScope` instance.
Might be better to replace those 2 params with a single `ValidatedDiscoveryScope` param, otherwise someone could theoretically pass a `discovery` and `scope_repr` that don't belong together.

chord_metadata_service/phenopackets/summaries.py
```diff
@@ -77,8 +81,20 @@ async def phenotypic_feature_summary(
...
+
+    # Start with either all phenopackets or a subset specified by a parameter
+    phenopackets = queryset if queryset is not None else models.Phenopacket.objects.all()
+
+    # Apply scope to existing queryset to enforce it on the summarization
+    if dataset_id := scope.dataset_id:
+        phenopackets = phenopackets.filter(dataset_id=dataset_id)
+    elif project_id := scope.project_id:
+        # Project is set but dataset isn't
+        phenopackets = phenopackets.select_related("project").filter(dataset__project_id=project_id)
```
v-rocheleau: Shouldn't this be `select_related("dataset__project")` ?

chord_metadata_service/experiments/summaries.py
```diff
@@ -102,8 +103,20 @@ async def instrument_summary(
...
+
+    # Start with either all experiments or a subset specified by a parameter
+    experiments = queryset if queryset is not None else models.Experiment.objects.all()
+
+    # Apply scope to existing queryset to enforce it on the summarization
+    if dataset_id := scope.dataset_id:
+        experiments = experiments.filter(dataset_id=dataset_id)
+    elif project_id := scope.project_id:
+        # Project is set but dataset isn't
+        experiments = experiments.select_related("project").filter(dataset__project_id=project_id)
```
v-rocheleau: Shouldn't this be `select_related("dataset__project")`?

v-rocheleau (changes requested):
Small questions, otherwise looks good!

chord_metadata_service/phenopackets/summaries.py
davidlougheed: should be `dataset` i think, but good point

chord_metadata_service/patients/api_views.py
davidlougheed: this was to avoid having to deal with `EmptyConfig` type in two places, since scopes can return empty configs. but yeah, i agree this also isn't the best (and the comment is truncated/unfinished)

v-rocheleau (approved):
LGTM!

PR #531, opened 2024-09-03
refact(discovery)!: introduce container class for discovery scope

davidlougheed:
goal of this is to encapsulate some of the discovery scope data in one place, and make it clearer at which point validation of the project/dataset ID happens.

non-UUID IDs for datasets/projects are now 404 (albeit can never happen) instead of 400 - the reason for changing this is that I figure the user shouldn't care about the ID beyond it being an opaque string, so having the same response for not-found/possible IDs makes more sense.

### DIFF
--- a/chord_metadata_service/chord/views_search.py
+++ b/chord_metadata_service/chord/views_search.py
@@ -13,6 +13,7 @@ from django.db.models import Count, F, Q
 from django.db.models.functions import Coalesce
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from psycopg2 import sql
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.permissions import AllowAny
@@ -528,8 +529,13 @@ DATASET_DATA_TYPE_SUMMARY_FUNCTIONS = {
 @async_api_view(["GET"])
 @permission_classes([BentoAllowAny])
 async def dataset_summary(request: DrfRequest, dataset_id: str):
-    dataset = await Dataset.objects.aget(identifier=dataset_id)
+    try:
+        dataset = await Dataset.objects.aget(identifier=dataset_id)
+    except (Dataset.DoesNotExist, ValidationError) as e:
+        return Response(errors.not_found_error(e.message), status=status.HTTP_404_NOT_FOUND)
+
     project = await Project.objects.aget(identifier=dataset.project_id)
+
     # don't use request scope - the project/dataset are validated by the aget calls above and fixed
     discovery_scope = ValidatedDiscoveryScope(project, dataset)
     dt_permissions = await get_data_type_query_permissions(
--- a/chord_metadata_service/chord/views_search.py
+++ b/chord_metadata_service/chord/views_search.py
@@ -487,18 +487,6 @@ def chord_dataset_search(
     return serialized_data, None
 
 
-def chord_dataset_representation(dataset: Dataset):
-    return {
-        "id": dataset.identifier,
-        "title": dataset.title,
-        "metadata": {
-            "project_id": dataset.project_id,
-            "created": dataset.created.isoformat(),
-            "updated": dataset.updated.isoformat(),
-        },
-    }
-
-
 def dataset_search(request: DrfRequest, dataset_id: str, internal=False):
     start = datetime.now()
 
--- a/chord_metadata_service/chord/views_search.py
+++ b/chord_metadata_service/chord/views_search.py
@@ -4,7 +4,6 @@ import json
 import logging
 
 from adrf.decorators import api_view as async_api_view
-from bento_lib.auth.resources import build_resource
 from bento_lib.responses import errors
 from bento_lib.search import build_search_response, postgres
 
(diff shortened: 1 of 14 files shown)


chord_metadata_service/discovery/utils.py
```diff
@@ -10,68 +12,150 @@
...
+    def __init__(self, project: cm.Project | None, dataset: cm.Dataset | None):
+        """
+        Constructor for an already-validated discovery scope - i.e., since we are getting fed project/dataset instances
+        rather than just string IDs, we know these objects exist at the time of construction.
+        """
+
+        self._project = project
+        self._dataset = dataset
+
+        # Additional validation - make sure we have project set if dataset is set
```
v-rocheleau: ~~Should also raise a `DiscoveryScopeException` if the project doesn't contain the dataset.~~

v-rocheleau: I see that `get_discovery_scope` checks the scope's validity before creating an instance of `ValidatedDiscoveryScope`, and that constructor usage is preceded by validation, but someone could still use the constructor to create an "invalid" `ValidatedDiscoveryScope` instance, which feels a bit off to me.

chord_metadata_service/discovery/utils.py
```diff
@@ -10,68 +12,150 @@
...
+            dataset = await qs.aget()
+            project = await _get_project_by_id(dataset.project_id)
+
         elif project_id:
-            # get project's discovery config if project_id is passed and dataset_id is not
-            return await _get_project_discovery(project_id=project_id)
+            project = await _get_project_by_id(project_id)
+
     except ObjectDoesNotExist:
-        raise DiscoveryConfigException(dataset_id, project_id)
```
v-rocheleau, line 63: We can simply raise a `DiscoveryScopeException` here and get rid of `is_scope_valid` since it is only used to raise that exception.

chord_metadata_service/chord/views_search.py
```diff
@@ -531,13 +529,16 @@ def private_dataset_search(request: DrfRequest, dataset_id: str):
 @permission_classes([BentoAllowAny])
 async def dataset_summary(request: DrfRequest, dataset_id: str):
     dataset = await Dataset.objects.aget(identifier=dataset_id)
```
v-rocheleau: try block for Dataset.DoesNotExist and ValidationError to return 404 status

chord_metadata_service/discovery/api_views.py
```diff
@@ -58,16 +54,18 @@ async def public_search_fields(request: DrfRequest):
     """
 
     try:
-        discovery = await get_request_discovery(request)
-    except DiscoveryConfigException as e:
+        discovery_scope = await get_request_discovery_scope(request)
+    except DiscoveryScopeException as e:
         return Response(e.message, status=status.HTTP_404_NOT_FOUND)
     except ValidationError as e:  # UUID error
         return Response(errors.bad_request_error(*e.messages), status=status.HTTP_400_BAD_REQUEST)
```
v-rocheleau: Shouldn't this be 404? Also `get_request_discovery_scope` catches UUID errors and raises a `DiscoveryScopeException`, so I don't think this except block is needed anymore.

chord_metadata_service/discovery/api_views.py
```diff
@@ -123,20 +121,23 @@ async def public_overview(request: DrfRequest):
     """
 
     try:
-        discovery = await get_request_discovery(request)
-    except DiscoveryConfigException as e:
+        discovery_scope = await get_request_discovery_scope(request)
+    except DiscoveryScopeException as e:
         return Response(e.message, status=status.HTTP_404_NOT_FOUND)
     except ValidationError as e:
         return Response(errors.bad_request_error(*e.messages), status=status.HTTP_400_BAD_REQUEST)
```
v-rocheleau: Idem

chord_metadata_service/discovery/tests/test_api.py
```diff
@@ -105,46 +116,58 @@ def assert_response_section_fields(self, response_obj: dict, config: dict):
             set(field for section in config["search"] for field in section["fields"])
         )
 
+    @staticmethod
+    def test_discovery_configs() -> TestDiscoveryConfigsDict:
```
v-rocheleau: The `test_` function prefix is usually reserved for functions that perform assertions.

chord_metadata_service/discovery/tests/test_api.py
```diff
@@ -105,46 +116,58 @@ def assert_response_section_fields(self, response_obj: dict, config: dict):
...
             # invalid UUID for project
-            ("counts", "?project=not-a-uuid", status.HTTP_400_BAD_REQUEST, None),
+            ("counts", "?project=not-a-uuid", status.HTTP_404_NOT_FOUND, "none"),
             # invalid UUID for dataset
-            ("counts", "?dataset=not-a-uuid", status.HTTP_400_BAD_REQUEST, None),
+            ("counts", "?dataset=not-a-uuid", status.HTTP_404_NOT_FOUND, "none"),
         ]
 
+        # use key aliases for configs to make subtest failure output more readable
+        tdc = self.test_discovery_configs()  # to get injected CONFIG_PUBLIC, need to calculate this in-test
```
v-rocheleau: Neat idea to inject this way!

v-rocheleau (changes requested):

chord_metadata_service/discovery/utils.py
v-rocheleau: Clarification: A user could do the following.
```python
p = cm.Project.objects.get(project_a_id)   # exists
d = cm.Dataset.objects.get(dataset_b_id)   # exists
discovery_scope = ValidatedDiscoveryScope(p, d)
```
If both `project_a` and `dataset_b` exist, but `dataset_b` doesn't belong to `project_a`, no exception will be raised even though the scope is technically invalid.

davidlougheed: should be fixed.

v-rocheleau (approved):
Lgtm

PR #535, opened 2024-09-09
chore!: clear out old CanDIG code

davidlougheed:
deleted ARGO code was for mCode -> ARGO conversion; since we've removed mCode, no reason to keep it


v-rocheleau (approved):
LGTM

PR #536, opened 2024-09-10
refact!: define a "scopeable model" base class + modelviewset scoping

davidlougheed:


davidlougheed:
@v-rocheleau this is still WIP, but I would like to know your thoughts on the general approach/abstractions chosen

gsfk:
is this not possible? 

`api/individuals?project=<id>`

davidlougheed:
should be - i'll double check...

chord_metadata_service/authz/viewset.py
```diff
@@ -0,0 +1,70 @@
...
+        else:
+            return None
+
+    async def request_has_data_type_permissions(
+            self, request: DrfRequest, scope: ValidatedDiscoveryScope | None = None
+    ):
+        # We MUST specifically mark view sets as scope-enabled (which means their queryset handles scope correctly);
+        # otherwise, we cannot scope into a specific project/dataset and must use the whole instance as the scope.
+        # Otherwise, we can could leak data from other projects/datasets.
+        # TODO: there must be a better way to enforce this without manual flagging
```
v-rocheleau: We could break this into two classes. One for unscoped authz, and another one that extends the first for scoped authz.

chord_metadata_service/chord/views_search.py
```diff
@@ -174,17 +191,23 @@ def search(request):
...
+    queryset = await query_function(scope, compiled_query, query_params, search_params)
 
     if search_params["output"] == OUTPUT_FORMAT_VALUES_LIST:
         result = {
             dataset_id: {
                 "data_type": data_type,
                 "matches": [p["value"] for p in dataset_dicts]
             } for dataset_id, dataset_dicts in itertools.groupby(
-                queryset,
+                [r async for r in queryset],
```
v-rocheleau: Why is this needed?

chord_metadata_service/experiments/models.py
```diff
@@ -78,7 +84,20 @@ def __str__(self):
...
+    @staticmethod
+    def get_scope_filters() -> ModelScopeFilters:
+        return {
+            "project": {
+                "filter": "experiment__dataset__project_id",
+                "prefetch_related": ("experiment__dataset",),
+            },
+            "dataset": {
+                "filter": "experiment__dataset_id",
+                "prefetch_related": ("experiemnt",),
```
v-rocheleau: Typo

chord_metadata_service/authz/viewset.py
```diff
@@ -0,0 +1,70 @@
...
+
+class BentoAuthzModelViewSet(
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
+    mixins.ListModelMixin,
+    BentoAuthzModelGenericViewSet
+):
+    pass
```
v-rocheleau: I would add a comment to indicate that this is the equivalent of a `viewsets.ModelViewSet` but with a custom `viewsets.GenericViewSet` implementation (`BentoAuthzModelGenericViewSet`).
I mistook it for a WIP piece of code before I looked into the changes in the views.

chord_metadata_service/discovery/scopeable_model.py
v-rocheleau, line 1: Good idea!

chord_metadata_service/experiments/models.py
```diff
@@ -17,7 +19,7 @@
 # model for the desired purposes.
 
 
-class Experiment(models.Model, IndexableMixin):
+class Experiment(BaseScopeableModel, IndexableMixin):
     """
     Class to store Experiment information. This model is primarily designed for genomic experiments; it is thus
     linked to a specific bisample.
```
v-rocheleau: Not a change, but we may as well fix this `bisample` typo here.

v-rocheleau (changes requested):
First pass

chord_metadata_service/experiments/api_views.py
```diff
@@ -47,30 +50,28 @@ class ExperimentViewSet(viewsets.ModelViewSet):
...
+            Experiment
+            .get_model_scoped_queryset(await get_request_discovery_scope(self.request))
+            .select_related(*EXPERIMENT_SELECT_REL)
+            .prefetch_related(*EXPERIMENT_PREFETCH)
+            .order_by("id")
+        )
 
 
-class ExperimentBatchViewSet(BatchViewSet):
+class ExperimentBatchViewSet(mixins.ListModelMixin, BentoAuthzModelGenericViewSet):
```
v-rocheleau: As it is now, all concrete views extending `BentoAuthzModelGenericViewSet` are also extending `mixins.ListModelMixin`.
If `BentoAuthzModelGenericViewSet` remains like that (an authz enabled replacement for BatchViewSet), I think it would be better to include the `mixins.ListModelMixin` in `BentoAuthzModelGenericViewSet`

chord_metadata_service/phenopackets/api_views.py
```diff
@@ -64,25 +85,44 @@ class BiosampleBatchViewSet(PhenopacketsModelViewSet):
         IndividualBentoSearchRenderer,
     )
     content_negotiation_class = FormatInPostContentNegotiation
+    permission_classes = (BentoDeferToHandler,)
```
v-rocheleau: Missing `scope_enabled` ?
Also, I think we can avoid using `BentoDeferToHandler` as the permission class and factor out the authz logic to `BentoAuthzModelGenericViewSet`.
Maybe by breaking `BentoAuthzModelGenericViewSet` into two classes, including one for batch viewsets

v-rocheleau (changes requested):

gsfk:
Search working well, only trouble so far is from `api/overview` which seems to ignore scope.

But I'll see if I can cobble something together from one of the /public endpoints instead.

chord_metadata_service/chord/views_search.py
davidlougheed: refactored + added a comment - queryset becomes an async iterator in an async context, and groupby only takes normal sync iterators/iterables.

chord_metadata_service/authz/viewset.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class BentoAuthzScopedModelGenericListViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
+    """
+    An extension of the DRF generic viewset which adds utility functions for Bento Django permissions classes.
+    These work together to properly implement scoped Bento permissions based on the request being made.
+
+    <!!!>
+    Security note: Subclasses MUST implement a get_queryset(...) which returns a model-scoped, request-based queryset!
+    </!!!>
```
v-rocheleau, line 25: We could enforce this by implementing a `get_queryset` here that raises a `NotImplementedError`.

chord_metadata_service/experiments/api_views.py
```diff
@@ -84,28 +86,35 @@ class ExperimentBatchViewSet(BatchViewSet):
...
+
+        return queryset.select_related(*EXPERIMENT_SELECT_REL).prefetch_related(*EXPERIMENT_PREFETCH).order_by("id")
+
+    @async_to_sync
+    async def get_queryset(self):
+        return self._get_filtered_queryset(self.request.data.get("id", None))
+
+    def permission_from_request(self, request: DrfRequest):
+        if self.action in ("list", "create"):
+            return P_QUERY_DATA
```
v-rocheleau: Would add a comment to specify that `create` maps to `P_QUERY_DATA` because `POST` requests are used to include body params for search.

chord_metadata_service/patients/models.py
```diff
@@ -21,13 +23,26 @@ class VitalStatus(BaseTimeStamp, IndexableMixin):
...
     """ Class to store demographic information about an Individual (Patient) """
 
     @property
     def schema_type(self) -> SchemaType:
         return SchemaType.INDIVIDUAL
 
+    @staticmethod
+    def get_scope_filters() -> ModelScopeFilters:
+        return {
+            "project": {
```
v-rocheleau, line 36:
```python
"filter": "phenopackets__dataset__project_id",
"prefetch_related": ("phenopackets__dataset_set",)
```

chord_metadata_service/phenopackets/api_views.py
```diff
@@ -65,24 +83,34 @@ class BiosampleBatchViewSet(PhenopacketsModelViewSet):
...
-        return queryset
+        return queryset.prefetch_related(*BIOSAMPLE_PREFETCH).order_by("id")
 
     def get_queryset(self):
-        individual_ids = self.request.data.get("id", None)
-        return self._get_filtered_queryset(ids_list=individual_ids)
+        return self._get_filtered_queryset(ids_list=self.request.data.get("id", None))
+
+    def permission_from_request(self, request: DrfRequest):
+        if self.action in ("list", "create"):
```
v-rocheleau, line 103: Would add a comment to explain why `create` maps to `P_QUERY_DATA`

chord_metadata_service/restapi/api_views.py
```diff
@@ -101,29 +52,41 @@ async def search_overview(request: DrfRequest):
     Overview statistics of a list of patients (associated with a search result)
     - Parameter
         - id: a list of patient ids
+        - project (optional), dataset (optional): scope for search overview
```
v-rocheleau, line 56: I think the permission_classes are missing for this endpoint.

chord_metadata_service/resources/tests/test_api.py
```diff
@@ -1,23 +1,137 @@
...
+        res = self.one_authz_get(self.url)
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(res.json()["results"]), 2)
+
+        # check that we don't have any resources under the project (until later tests...)
+        res = self.one_authz_get(self.url_with_proj)
+        self.assertEqual(res.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(res.json()["results"]), 0)
+
+    def test_list_resources_scope_dne(self):
```
v-rocheleau, line 72: Not sure I understand why these are expecting 403 responses with `one_authz_get`

chord_metadata_service/authz/viewset.py
v-rocheleau, line 1: Would it be too complicated to have full test cov here?

v-rocheleau (changes requested):
Final requests I think, looks good!

chord_metadata_service/restapi/api_views.py
davidlougheed: the default is BentoDeferToHandler (defined in settings), but I'll make it explicit

gsfk (approved):
Working well for me so far.

chord_metadata_service/authz/viewset.py
davidlougheed: was being lazy, done.

chord_metadata_service/resources/tests/test_api.py
v-rocheleau: Explanation: 403 happens with an invalid scope because the permissions cannot be determined on an authz protected endpoint.
We can add a note in a comment for that.

gsfk:
post-approval comment: behaviour when mismatching projects & datasets is not always consistent. If I manage to request a project_id and and an incorrect dataset_id (from another project, say), I get a mix of behaviours: 

- `/api/public_rules` returns a sensible error message
- `/api/individuals` returns "Forbidden"

working now, thanks

chord_metadata_service/restapi/exception_handler.py
v-rocheleau, line 1: Nice!

v-rocheleau (approved):
LGTM! Nice work

PR #538, opened 2024-09-11
fix(chord): typo in dict comprehension leading to bad request

davidlougheed:
was supposed to be a dictionary of {data type: summary}, ended up accidentally as [{data type 1: sumary}, {data type 2: summary}]
also adds a test for response format for the dataset summary endpoint to prevent this from happening again in the future.

should fix web dataset data type summaries in manager.


v-rocheleau (approved):
lgtm

PR #540, opened 2024-09-16
chore: publish json schemas as release artifacts

v-rocheleau:
- new django manage.py script `schemas` to generate single file JSON-schemas for phenopackets, experiments and discovery
- new `release` github actions workflow is triggered on new releases and adds the schemas to the release as an artifact zip file

Tested with a [dummy](https://github.com/bento-platform/katsu/releases/tag/v8.0.2-test) pre-release with a PR trigger. 
The content of the `json-schemas-clean.zip` artifact is what it will look like on a real release, but with the `json-schemas.zip` name (artifacts cannot be overriden, so I changed the names as I was testing to avoid errors).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,3 +33,16 @@ jobs:
           image-name: ghcr.io/bento-platform/katsu
           development-dockerfile: bento.dev.Dockerfile
           dockerfile: bento.Dockerfile
+
+      # TODO: test, maybe just add to a protected build directory?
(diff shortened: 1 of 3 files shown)


davidlougheed:
you can probably fix the errors right now by setting up python, running `poetry install` before export, and doing `poetry run ./manage.py [...]` in order to use the poetry env to run the script.

davidlougheed (approved):
lgtm

PR #541, opened 2024-09-17
chore!: remove Django admin interface (was debug only)

davidlougheed:
This isn't used and provides a potential security hole, and probably will not play nice with the Bento authorization scheme.


v-rocheleau (approved):
lgtm

PR #542, opened 2024-09-17
fix: proper permissions for schema endpoints

davidlougheed:
these should be properly public.


v-rocheleau (approved):
Lgtm. There are still decorators using the `AllowAny` permission class, should they be replaced as well?

davidlougheed:
There are a lot of endpoints (search mostly) I haven't thought through yet, so for now I'm leaving them as-is - these were already included in the authz middleware allow-list, but didn't have the proper permissions to be served.

PR #543, opened 2024-09-19
fix(actions): single release workflow trigger for pre-release and stable

v-rocheleau:
The `published` release type is triggered for both stable and pre-release.
Having both `published` and `prerelease` caused the workflow to run in double on pre-releases.


davidlougheed (approved):
lgtm

PR #544, opened 2024-09-27
fix(chord): fix bad logic for checking if dataset project changed

davidlougheed:
adds a test which catches this (and fixes some assert patterns in that test file.)


gsfk (approved):
works as expected, thanks (I didn't look at the tests)

PR #553, opened 2024-11-27
chore!: remove unused endpoints + FHIR + Elastic

davidlougheed:


davidlougheed:
@gsfk could you verify i'm not removing anything you use here?

gsfk (approved):

chord_metadata_service/patients/models.py
```diff
@@ -71,10 +71,6 @@ def get_project_id(self) -> str | None:
     gender = JSONField(blank=True, null=True, validators=[ontology_validator],
                        help_text='Self-identified gender')
 
-    # FHIR specific
-    active = models.BooleanField(default=False, help_text='Whether this patient\'s record is in active use.')
-    deceased = models.BooleanField(default=False, help_text='Indicates if the individual is deceased or not.')
-
```
v-rocheleau, line 77: These fields are still in the Individual schema and descriptions dict.

v-rocheleau (changes requested):
Small request, will take a closer look at interpretation changes first thing tomorrow.

chord_metadata_service/patients/models.py
davidlougheed: good point

v-rocheleau (changes requested):
Looks good, validated that the endpoints removed are not used elsewhere.
Few things:
- There is still the (now unused) `fhir_utils.py` file, which links to hl7 and FHIR mapping files.
- `restapi/schemas` has an unused `FHIR_BUNDLE_SCHEMA`
- `restapi/semantic_mappings/README.md` can probably be removed

v-rocheleau (commented):
Lint fails, otherwise LGTM!

PR #556, opened 2024-12-09
refact: single experiment results file ingestion workflow

davidlougheed:
this will be useful for ingesting any type of experiment result (for, e.g., RENATA)


noctillion (approved):
LGTM! It works as intented.

PR #564, opened 2025-01-23
feat(discovery)!: handle boolean permissions for rules/overview

davidlougheed:
mostly for Beacon.


gsfk:
working for me

gsfk (approved):
Still working as intended for me. 

Are there response changes for boolean queries? Beacon typically won't see these since it usually searches with full auth.

PR #565, opened 2025-01-24
chore: add missing migrations (seem to be JSON field migrations)

davidlougheed:


v-rocheleau (approved):
The diffs come from JSON schema changes.

We can proceed with this to get 18 out, but I think we should add guardrails for this.

In order to avoid missing migrations for future releases, we should look into incorporating the  `manage.py makemigrations --check` [command](https://docs.djangoproject.com/en/5.1/ref/django-admin/#cmdoption-makemigrations-check) in our GH actions.

For productivity I think it would be best if we have this triggered before releases only, so we don't have to create multiple feature migrations that we need to squash later for the release.

We could have Katsu to adopt the `releases/<version>` branch pattern, and we run this check on PRs from `releases/<version>` to `develop`.

PR #568, opened 2025-01-29
feat: add biosample collection_location + GeoLocation model

davidlougheed:


davidlougheed:
still need to write a few more tests, but requesting review.

chord_metadata_service/geo/models.py
```diff
@@ -0,0 +1,59 @@
...
+        blank=True,
+        default=dict,
+        validators=[base_extra_properties_validator],
+        help_text="Extra properties that do not have a predefined field in the database.",
+    )
+
+    # ------------------------------------------------------------------------------------------------------------------
+
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
```
v-rocheleau: Extend `chord_metadata_service.restapi.models.BaseTimeStamp` instead.

chord_metadata_service/geo/models.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    precision = models.TextField(blank=True, default="", help_text=d.PROP_ISO3166_ALPHA_3)
+
+    # properties (mapping to other properties in the GeoJSON object) which do not map to any of the above fields:
+    extra_properties = models.JSONField(
+        blank=True,
+        default=dict,
+        validators=[base_extra_properties_validator],
+        help_text="Extra properties that do not have a predefined field in the database.",
+    )
```
v-rocheleau, line 48: Could use `chord_metadata_service.restapi.models.BaseExtraProperties` instead, if we want to specify JSON schemas for the `GeoLocation.extra_properties` field. 
Given how much Genovalia seem to enjoy the extra properties schemas, I think they will find this option useful here as well.
Since Biosample already extends it and `GeoLocation` is a member of a `Biosample`, we can easily implement the `BaseExtraProperties.get_project_id` function.

chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -141,6 +143,25 @@ def update_or_create_subject(subject: dict) -> pm.Individual:
     return subject_obj
 
 
+def get_or_create_geo_location(geoloc_json: dict) -> gm.GeoLocation:
```
v-rocheleau: I need to read more on PostGIS and Django GIS floating point precision.
I'm worried that get_or_create will have issues with very close points, would be worth to test the behavior IMO.
One thing we can do is round the input coordinates to the degree of precision we expect, 7 decimals should be enough for our application.

chord_metadata_service/geo/models.py
```diff
@@ -0,0 +1,59 @@
...
+    city = models.TextField(blank=True, default="", help_text=d.PROP_CITY)
+    country = models.TextField(blank=True, default="", help_text=d.PROP_COUNTRY)
+    iso_a3_code = models.CharField(
+        max_length=3,
+        choices=ISO_3166_1_ALPHA_3_COUNTRY_CODE_CHOICES,
+        null=True,
+        default=None,
+        help_text=d.PROP_ISO3166_ALPHA_3,
+    )
+    precision = models.TextField(blank=True, default="", help_text=d.PROP_ISO3166_ALPHA_3)
```
v-rocheleau: Wrong `help_text`

v-rocheleau (changes requested):

chord_metadata_service/geo/models.py
davidlougheed: one thing I was worried about with schemas was if the model is used in two different places (e.g., biosample + dataset) the schemas become slightly different - but I guess `oneOf` mostly takes care of this. anyway, sounds good overall.

chord_metadata_service/chord/ingest/phenopackets.py
davidlougheed: image

chord_metadata_service/geo/models.py
davidlougheed: actually, `get_project_id` becomes hard if we use this in multiple places. can we leave this for later perhaps?

chord_metadata_service/geo/models.py
davidlougheed: done

chord_metadata_service/geo/models.py
davidlougheed: fixed

chord_metadata_service/chord/ingest/phenopackets.py
v-rocheleau: We can keep this vanilla for now and adapt the way we validate/clean coordinates as we go.

chord_metadata_service/geo/models.py
v-rocheleau: We can leave this for later, didn't think of dataset.
Since biosamples can have extra_properties schemas, and a biosample can only have 1 GeoLocation, we can still add/validate relevant collection information at the `Biosample.extra_properties` level.

chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -141,6 +143,25 @@ def update_or_create_subject(subject: dict) -> pm.Individual:
...
+    """
+    gs.GeoLocationSerializer(data=geoloc_json).is_valid(raise_exception=True)
+    geoloc, _ = gm.GeoLocation.objects.get_or_create(
+        point=Point(geoloc_json["geometry"]["coordinates"]),
+        **{
+            gc.MODEL_PREDEF_PROPS_TO_ATTRS[gk]: gv
+            for gk, gv in geoloc_json.get("properties", {}).items()
+            if gk in gc.MODEL_PREDEF_PROPS_TO_ATTRS
+        },
+        # TODO: extra properties
```
v-rocheleau: Still TODO?

chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -141,6 +143,25 @@ def update_or_create_subject(subject: dict) -> pm.Individual:
...
 
+def get_or_create_geo_location(geoloc_json: dict) -> gm.GeoLocation:
+    """
+    Given a GeoJSON-based representation of a location, return an existing (re-used) or new GeoLocation object, now
+    saved in the database. The provided dictionary should be compatible with the Katsu GeoLocation format, based on the
+    GA4GH schema block / Progenetix GeoLocation schema< https://schemablocks.org/schema_pages/Progenetix/GeoLocation/>.
+    """
+    gs.GeoLocationSerializer(data=geoloc_json).is_valid(raise_exception=True)
+    geoloc, _ = gm.GeoLocation.objects.get_or_create(
+        point=Point(geoloc_json["geometry"]["coordinates"]),
```
v-rocheleau: `geos.Point` can represent coordinates in different representations formats, while GeoJSON represents all coordinates with SRID 4326. So we should specify it when creating the `Point` here, since we can assume all our inputs are in that representation.

v-rocheleau (changes requested):

chord_metadata_service/chord/ingest/phenopackets.py
davidlougheed: done!

chord_metadata_service/chord/ingest/phenopackets.py
davidlougheed: oops - finished this, added tests for it too

davidlougheed:
@v-rocheleau found some issues with updates and permissions too, now fixed.

v-rocheleau (approved):
LGTM!

PR #572, opened 2025-03-24
indexing alignment files on ingestion

noctillion:
Index cram and bam files if the index is not available on ingestion.

### DIFF
--- a/chord_metadata_service/chord/workflows/wdls/experiment_results_files.wdl
+++ b/chord_metadata_service/chord/workflows/wdls/experiment_results_files.wdl
@@ -9,14 +9,14 @@ workflow experiment_results_files {
         Boolean validate_ssl
     }
 
-    scatter(file in files) {
+    scatter (file in files) {
         call post_to_drs {
             input:
-                file_path = file,
-                drs_url = drs_url,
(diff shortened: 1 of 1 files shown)


chord_metadata_service/chord/workflows/wdls/experiment_results_files.wdl
```diff
@@ -33,19 +33,61 @@ task post_to_drs {
...
     command <<<
+        # Handling index files
+        handle_index_file() {
+            local file_extension="$1"
+            local index_extension="$2"
+
+            index_file="~{file_path}${index_extension}"
+            index_name=$(basename "$index_file")
+
+            # Search for the index file in DRS
```
davidlougheed, line 46: I think it makes more sense to check if the index file is listed in the experiment results JSON during the experiment+results workflow rather than this workflow, and if so ingest it.

I'm not sure if this is the right workflow to modify, or if you modify this one as well, maybe just check if the index file is included in the file array and if not, generate the index. Relying on filenames like you do here is not good practice with DRS since the only true identifiers are the DRS ID / hash.

davidlougheed (changes requested):

PR #573, opened 2025-04-01
feat: enable authz for projects and datasets lists

v-rocheleau:
Enables optional authorization checks for project and dataset models, mostly intended for the data catalog.

- `GET /projects`
   - User must have a grant with the `P_VIEW_PROJECTS` permission
   - Feature flag: `KATSU_PROJECTS_LIST_AUTHZ: bool`
- `GET /datasets`
   - User must have a grant with the `P_VIEW_DATASETS` permission
   - Feature flag: `KATSU_DATASETS_LIST_AUTHZ: bool`

### DIFF
--- a/chord_metadata_service/chord/api_views.py
+++ b/chord_metadata_service/chord/api_views.py
@@ -79,7 +79,7 @@ class ProjectViewSet(CHORDPublicModelViewSet):
     Create a new project
     """
 
-    queryset = Project.objects.all().order_by("identifier")
+    queryset = Project.objects.all().order_by("title")
     serializer_class = ProjectSerializer
 
     @async_to_sync
--- a/chord_metadata_service/chord/api_views.py
+++ b/chord_metadata_service/chord/api_views.py
@@ -9,6 +9,8 @@ from bento_lib.auth.permissions import (
     P_CREATE_DATASET,
     P_EDIT_DATASET,
     P_DELETE_DATASET,
+    P_VIEW_PROJECTS,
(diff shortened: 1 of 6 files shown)


chord_metadata_service/chord/api_views.py
```diff
@@ -117,6 +120,15 @@ async def destroy(self, request, *args, **kwargs):
         authz.mark_authz_done(request)
         return await sync_to_async(super().destroy)(request, *args, **kwargs)
 
+    @async_to_sync
+    async def list(self, request, *args, **kwargs):
```
davidlougheed, line 124: there are a few more endpoints for viewing project/dataset metadata (a bunch of public endpoints + maybe some old CHORD endpoints too), I think. Also, how will this interface with Beacon?

chord_metadata_service/chord/api_views.py
```diff
@@ -25,6 +27,7 @@
 from chord_metadata_service.authz.middleware import authz_middleware as authz
 from chord_metadata_service.authz.permissions import BentoAllowAnyReadOnly, BentoDeferToHandler
 from chord_metadata_service.cleanup.run_all import run_all_cleanup
+from chord_metadata_service.metadata.settings import KATSU_DATASETS_LIST_AUTHZ, KATSU_PROJECTS_LIST_AUTHZ
```
davidlougheed, line 30: use `from django.conf import settings` and access all Django settings thru that interface

davidlougheed (changes requested):

PR #574, opened 2025-04-07
feat(ingest): index alignment files on experiment JSON ingest

noctillion:
Index cram and bam files if the index is not available on ingestion in the experiments_json_with_files workflow.


chord_metadata_service/chord/workflows/wdls/experiments_json_with_files.wdl
```diff
@@ -146,20 +146,47 @@ task post_to_drs {
...
             -F "file=@~{file_path}" \
             -F "project_id=$project_id" \
             -F "dataset_id=$dataset_id" \
             -H "Authorization: Bearer ~{token}" \
             --fail-with-body \
-            "~{drs_url}/ingest"
+            "~{drs_url}/ingest")
+        echo "$resp_main" | jq -c
+
+        # If it's BAM or CRAM, ingest the index as well
```
davidlougheed: this part of the code still isn't matching desired behaviour exactly - indices should be ingested if they're listed in the JSON; or, if no index is specified in the JSON, one should be generated (maybe in the python script using subprocess?) and put into the JSON. I think also `prepare_for_drs` should extract index files as well as file paths, and this task should stay naive to the type of file it is ingesting.

davidlougheed: I think for now the best way to write this workflow might unfortunately just be as a single big Python task which uses the `requests` library to ingest into DRS and, if needed, uses subprocess+samtools to generate indices for files missing the indices, rewriting the experiments JSON as it goes and then POSTing it to Katsu at the end.

davidlougheed (changes requested):

davidlougheed (approved):
works for me.

PR #577, opened 2025-04-14
feat!: new discovery API

davidlougheed:
corresponding bento public PR: https://github.com/bento-platform/bento_public/pull/270


chord_metadata_service/chord/data_types.py
```diff
@@ -1,17 +1,21 @@
...
     "DATA_TYPES",
 ]
 
 
-DATA_TYPE_EXPERIMENT = "experiment"
-DATA_TYPE_PHENOPACKET = "phenopacket"
+KatsuDataType = NewType("KatsuDataType", str)
+
+DATA_TYPE_EXPERIMENT = KatsuDataType("experiment")
+DATA_TYPE_PHENOPACKET = KatsuDataType("phenopacket")
```
v-rocheleau, line 20: Was not familiar with NewType, did this change uncover bad type checks?

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
             # key to use across all discovery endpoints:
-            count_or_bools_res[public_model_name] = model_count if model_permissions["counts"] else (model_count > 0)
+            count_or_bools_res[e] = entity_count if entity_permissions.counts else (entity_count > 0)
 
-    # Parse the public config to gather data for each field defined in the overview
+    if (
+        not count_or_bools_res[queryset_model_name]
+        and not dt_permissions[DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE[queryset_model_name]].data
+    ):
+        message = INSUFFICIENT_DATA_AVAILABLE_MSG
```
v-rocheleau, line 385: If we go into this block, we still return the counts in a `DiscoveryResponse` at the end of the func.
To me this feels like it should be an early non 200 return? I guess you have that planned already from the TODO comment bellow, just wondering

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
+
+    dt_permissions = await get_discovery_data_type_permissions(request, scope)
+    if not dt_permissions[dts.DATA_TYPE_PHENOPACKET].data:
+        # "Extra" permissions check vs. regular discovery endpoint: we need full data permissions
+        return dres.insufficient_privileges(request)
+    authz_middleware.mark_authz_done(request)
+
+    # -- Query execution -----------------------------------------------------------------------------------------------
+
+    queried_entity: DiscoveryEntity = request.query_params.get("_entity", "phenopacket")
```
v-rocheleau, line 452: Just to help me understand what the query parameter naming logic is here.
I think having `_` prefixes makes sense for generic stuff like pagination options, but the query entity is core to discovery so it surprised me at first.
If the goal is to reserve non-prefixed query params for queriable fields I like the idea, but we should write this somewhere I think.

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
-            name='public_overview_not_available',
-            fields={'message': serializers.CharField()},
-        ),
-    }
-)
-@api_view(["GET"])
+    return {**counts, "individual": len(individual_id_set)}
+
+
+@api_view(["GET", "POST"])
```
v-rocheleau: Would add a comment to specify when GET or POST should be used

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
+            message=message,
+            # permissions-dependent: dictionary of {entity: counts or True if above threshold, 0/False otherwise}:
+            counts=count_or_bools_res,
+        )
+    )
+
+
+@api_view(["GET", "POST"])
+@permission_classes([BentoDeferToHandler])
+async def discovery_matches(request: DrfRequest):
```
v-rocheleau, line 408: Would add a comment to specify when GET or POST should be used

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
+        return await dres.django_validation_error(
+            request, e, lg, "discovery matches endpoint recieved validation error"
+        )
+
+    lg = lg.bind(query=query.model_dump(mode="json"))
+
+    # -- Pagination ----------------------------------------------------------------------------------------------------
+
+    page: int = int(request.query_params.get("_page", "0"))
+    page_size = int(request.query_params.get("_page_size", "10"))  # if page_size is set to 0,
```
v-rocheleau: We use a default page_size of 25 in `LargeResultsSetPagination`, I think that 10 is pretty small.
Would also be good to validate the signs of `_page` and `_page_size`, as well as a max size.

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
+
+@api_view(["GET"])
+@permission_classes([BentoAllowAny])
+async def discovery_ui_hints(request: DrfRequest):
+    try:
+        scope = await get_request_discovery_scope(request)
+    except DiscoveryScopeException as e:
+        return Response(errors.not_found_error(e.message), status=status.HTTP_404_NOT_FOUND)
+
+    queryset_model_name: DiscoveryEntity = "phenopacket"  # TODO
```
v-rocheleau: What is the todo for here?

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,225 +1,437 @@
...
+    dt_permissions = await get_discovery_data_type_permissions(request, scope)
+
+    return {
+        "entities_with_data": [
+            e for e, v in counts.items()
+            if thresholded_count(v, scope, dt_permissions[DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE[e]])
+        ],
+        # TODO: instead of this, maybe we also collect experiment results and check for geojson, and indicate if we
+        #  should present a consolidated map view?
+        "biosample_location_present": False,  # TODO: non-Null location_collected above threshold
```
v-rocheleau: Think we should comment this out if this is out of scope for this PR.
What do you mean by a consolidated map view? 
Don't know if thats what you have already in mind, but it would be great if clients could retrieve GeoJSON FeatureCollections grouped for the current scope, so that we can render it directly in leaflet without client-side processing.

v-rocheleau (changes requested):
First pass, I like the structlog scope contexts you added!

chord_metadata_service/discovery/api_views.py
davidlougheed: yes, it's so that we can have easy logic to grab query-able fields and avoid any namespace conflicts with discovery config (e.g., if someone named a field in their config "entity"). I'll write a note on that.

chord_metadata_service/discovery/api_views.py
davidlougheed: i think we only need GET

chord_metadata_service/discovery/api_views.py
davidlougheed: i think we only need GET

chord_metadata_service/discovery/api_views.py
davidlougheed: yeah this should be implemented later.

chord_metadata_service/discovery/api_views.py
davidlougheed: made it more specific.

chord_metadata_service/chord/data_types.py
davidlougheed: mostly done for clarity for where these are passed elsewhere to functions iirc

chord_metadata_service/discovery/api_views.py
davidlougheed: added a constant for this and added some validation, max_size up next

chord_metadata_service/discovery/api_views.py
davidlougheed: I'm pretty sure I did this so that we could still get layout data if a search narrowed to the point that we no longer had sufficient data, but I'll double check this.

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,246 +1,567 @@
...
+        )
+
+        return filtered_queryset, queried_entities
+
+    async def get_query_queryset_and_queried_entities(
+        self,
+        entity: DiscoveryEntity,
+        lg: BoundLogger | None = None,
+    ) -> tuple[QuerySet, frozenset[DiscoveryEntity]]:
+        async with self._queryset_locks[entity]:
```
v-rocheleau, line 132: Why do we need an asyncio Lock here? Just for my understanding

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,246 +1,567 @@
...
+        # We return None and this field presumably gets filtered out, with logs on the backend
+        # TODO: some type of field response error we can return to the front-end - a DiscoveryFieldResponse error prop?
+        return None
+
+    return DiscoveryFieldResponse(id=field, definition=field_props, data=stats)
+
+
+async def discovery_queryset_entity_counts(qqs: QueryQuerysetsCache) -> dict[DiscoveryEntity, int]:
+    return {
+        e: await (await qqs.get_query_queryset_and_queried_entities(e))[0].acount()
```
v-rocheleau: Would add a comment here. I was not sure why the [0] was there at first, but looking at the QQS returns it made sense. Maybe just mention that it is a tuple whose first element is always a queryset

chord_metadata_service/discovery/api_views.py
```diff
@@ -1,246 +1,567 @@
...
+        and not dt_permissions[DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE[queryset_entity]].data
+    ):
+        message = INSUFFICIENT_DATA_AVAILABLE_MSG
 
-    fields = discovery.get_chart_field_ids()
-    fields_res: dict[str, dict] = {}
-    _, field_permissions = get_discovery_field_set_permissions(discovery, fields, dt_permissions)
+    # If phenopacket is 0, don't reveal nested entities exist, otherwise we could get responses like (in the case of
+    # one phenopacket with five biosamples): { phenopacket: 0, biosample: 5, ... }
+    # TODO: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
v-rocheleau: TODO for this PR or later? Asking because of the amount of exclamation points haha

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -30,36 +49,233 @@ def get_jsonb_path_query(field: str, json_path: str, is_array=True, is_mapping=T
...
+        case ("experiment_result", "experiment"):
+            if field_path[:1] == ("experiment_result",):
+                return field_path[1:]
+            return "experiments", *field_path
+        # --------------------------------------------------------------------------------------------------------------
+        case _:
+            raise exc
+
+
+field_path_to_django_mapping = "__".join
```
v-rocheleau: Move with other constants

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -30,36 +49,233 @@ def get_jsonb_path_query(field: str, json_path: str, is_array=True, is_mapping=T
     return JSONBPathQuery(F(field), Value(f"{field_operator}.{query_path}"))
 
 
-def get_public_model_name_and_field_path(field_id: str) -> tuple[str, tuple[str, ...]]:
-    model_name, *field_path = field_id.split("/")
-    return model_name, tuple(field_path)
+# noinspection PyUnreachableCode
```
v-rocheleau: Is the noinspection necessary here?

chord_metadata_service/discovery/fields_utils.py
```diff
@@ -30,36 +49,233 @@ def get_jsonb_path_query(field: str, json_path: str, is_array=True, is_mapping=T
...
+            return normalize_field_path_true_model("phenopacket", tuple(rest))
+        case ("individual", ("biosamples", *rest)):
+            # NOTE: this is a non-standard access pattern, but still works as of the time of writing (2025-09-03)
+            # because of the way the Django relationships are set up.
+            return normalize_field_path_true_model("biosample", tuple(rest))
+        case ("phenopacket", ("subject", *rest)):
+            return normalize_field_path_true_model("individual", tuple(rest))
+        case ("phenopacket", ("biosamples", *rest)):
+            return normalize_field_path_true_model("biosample", tuple(rest))
+        case ("biosample", ("experiment" | "experiments", *rest)):  # "experiment" is old name, "experiments" is new
```
v-rocheleau, line 310: May want to add a removal todo like you did for the other singular "experiment" field names

chord_metadata_service/discovery/pydantic_models.py
v-rocheleau, line 1: Really nice to have all that now, thanks!

chord_metadata_service/discovery/stats.py
```diff
@@ -79,56 +77,66 @@ async def bento_public_format_count_and_stats_list(
...
     Computes counts of distinct values for a queryset.
     """
 
+    # to prevent a JSONB path query from conflicting with a potentially real key on the queryset, we cannot use just the
+    # field access path as a unique ID - we can include the group_by clause as well (and add a prefix) to prevent a
+    # collision, e.g.:
+    #  with mapping=individual/phenopackets/medical_actions, group_by=procedure/code/label, just using mapping (field)
+    #  as the unique key would collide with the real medical_actions field on phenopackets after we normalize mapping.
+    #  By instead using _jsonb_medical_actions_procedure_code_label as the annotation key, we have something unique.
+    queryset_key = f"_jsonb_{field}_{group_by.replace(MAPPING_SEPARATOR, '_')}" if group_by is not None else field
```
v-rocheleau, line 119: Nice!

v-rocheleau (changes requested):
Not a full pass, hopping on Julian's PRs, will come back after.
Looks good, mostly questions

chord_metadata_service/discovery/fields_utils.py
davidlougheed: done

chord_metadata_service/discovery/fields_utils.py
davidlougheed: done

chord_metadata_service/discovery/fields_utils.py
davidlougheed: my ide was glitching out for a while, but yes this is removed now

chord_metadata_service/discovery/api_views.py
davidlougheed: added some explanatory comments

chord_metadata_service/discovery/api_views.py
davidlougheed: added an explanatory note

chord_metadata_service/discovery/api_views.py
davidlougheed: addressed this

chord_metadata_service/discovery/matches.py
v-rocheleau, line 1: **suggestion(non-blocking):** would be nice to have a top-level comment describing what matches are and how they relate to dicsovery

chord_metadata_service/experiments/filters.py
```diff
@@ -43,7 +43,7 @@ class ExperimentResultFilter(django_filters.rest_framework.FilterSet):
     # filter by datasets
     datasets = django_filters.CharFilter(
         method=filter_datasets,
-        field_name="experiment__dataset_id",
+        field_name="experiments__dataset_id",
```
v-rocheleau, line 46: **question:** should this also support the singular experiment notation for now?

chord_metadata_service/restapi/api_renderers.py
```diff
@@ -256,3 +260,24 @@ class IndividualBentoSearchRenderer(JSONRenderer):
...
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        return super().render(_json_dump_if_pyd(data), accepted_media_type, renderer_context)
+
+
+class PydanticBrowsableAPIRenderer(BrowsableAPIRenderer):
+    """
+    TODO
+    """
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        return super().render(_json_dump_if_pyd(data), accepted_media_type, renderer_context)
```
v-rocheleau, line 284: Neat, good idea!

v-rocheleau (approved):
LGTM! Small non-blocking comments

chord_metadata_service/experiments/filters.py
davidlougheed: this is an internal lookup, so it needs to correspond to the new name (It cannot have a fallback.)

chord_metadata_service/restapi/api_renderers.py
davidlougheed: thanks!

PR #579, opened 2025-04-28
feat(discovery): detailed matches for discovery individual response

davidlougheed:
adds a second matches list object for discovery endpoints to return project/dataset ID and phenopacket ID in addition to subject IDs. this is quasi-temporary, until we can get around to adding an improved search endpoint...


gsfk (approved):
Works well once I woke up and figured out how to search. 

Do we prefer flat results to ones that are divided by project / dataset? I offer no real opinion as to which is better.

davidlougheed:
not sure, but for now this is just left as mostly backwards-compatible - we can figure out how to divide stuff up for new search after.

PR #580, opened 2025-04-28
fix: serializer n+1 query issues

davidlougheed:
* fix n+1 query issues in CSV export
* add SQL query logging in debug mode


gsfk (commented):
Should I expect regular bento web csv export to get faster with this? It's still quite slow.

davidlougheed:
it should be faster... not necessarily fast.

gsfk (approved):
Seems to work as intended.

PR #581, opened 2025-05-06
refact(discovery)!: use singular public model name for counts

davidlougheed:


gsfk (approved):
Good, as long as
- we can keep pluralized endpoint names
- keep plurals in UI where reasonable
- make this consistent with other responses

PR #583, opened 2025-05-21
feat: add dataset `conditions_of_access` field

BarrothTheDragon:
**Task**                        
In katsu, add a field to datasets to store Data Access Request details, trying to be as close to schema.org as we can.


chord_metadata_service/chord/migrations/0010_dataset_conditionsofaccess_alter_dataset_created_and_more.py
v-rocheleau, line 1: Migrations include non-camel case that we wont use. In local:
1. Migrate to before these migrations using manage.py
2. Delete the new migrations
3. Regenerate the migrations
This way we won't have intermediate field migrations that were only used in dev.

chord_metadata_service/chord/migrations/0011_rename_conditionsofaccess_dataset_conditions_of_access.py
v-rocheleau, line 1: idem

chord_metadata_service/chord/models.py
```diff
@@ -89,6 +89,9 @@ class Dataset(BaseProjectOrDataset):
     linked_field_sets = models.JSONField(blank=True, default=list,
                                          help_text="Data type fields which are linked together.")
 
+    conditions_of_access = models.JSONField(blank=True, null=True,
```
v-rocheleau: Sorry I misled you in using a JSONField too fast, this should be a TextField per schema.org's [spec](https://schema.org/conditionsOfAccess)

chord_metadata_service/chord/tests/test_api.py
```diff
@@ -113,11 +113,14 @@ def setUp(self) -> None:
                 **valid_dataset_1(self.project["identifier"]),
                 "title": "Dataset 2",
                 "dats_file": {},  # Valid dats_file JSON object
+                "conditions_of_access": {},
+
```
v-rocheleau: Lint: unnecessary line return

chord_metadata_service/chord/tests/test_api.py
```diff
@@ -167,6 +170,14 @@ def test_create_dataset_forbidden(self):
         r = self.one_no_authz_post("/api/datasets", json=self.valid_payloads[0])
         self.assertEqual(r.status_code, status.HTTP_403_FORBIDDEN)
 
+    def test_conditions_of_access(self):
+        for i, d in enumerate(self.valid_payloads):
+            r = self.one_authz_post("/api/datasets", json=d)
+            expected = d["conditions_of_access"] if i > 0 else None
```
v-rocheleau: We can determine if the expected `condition_of_access` directly from the payload itself, eliminating the need for the unreliable index condition (this test would no longer work if we add payload items). The `dict.get` [function](https://docs.python.org/3/library/stdtypes.html#dict.get) is pretty useful to handle situations where we know a key may not be present.
```python
# returns None by default if the 'conditions_of_access' key is not found in 'd'
expected  = d.get("conditions_of_access")
```

v-rocheleau (changes requested):
I'll look into why we have geolocation related migrations generated here, they should not be included in this PR.

chord_metadata_service/chord/models.py
```diff
@@ -89,6 +89,9 @@ class Dataset(BaseProjectOrDataset):
     linked_field_sets = models.JSONField(blank=True, default=list,
                                          help_text="Data type fields which are linked together.")
 
+    conditions_of_access = models.TextField(blank=True, null=True,
+                                            help_text="The data access requests link(s), in JSON format")
```
v-rocheleau: Change help text, this is in plain text format, also include ref to https://schema.org/conditionsOfAccess

v-rocheleau (changes requested):

chord_metadata_service/chord/models.py
```diff
@@ -89,6 +89,10 @@ class Dataset(BaseProjectOrDataset):
     linked_field_sets = models.JSONField(blank=True, default=list,
                                          help_text="Data type fields which are linked together.")
 
+    conditions_of_access = models.TextField(blank=True, null=True,
```
davidlougheed: I think we should not let this be null, but instead treat blank strings as "empty", to avoid two possible empty values; then we could set the default to `""`. @v-rocheleau thoughts on this PR? it's been open for a while

chord_metadata_service/chord/tests/test_api.py
```diff
@@ -128,7 +130,8 @@ def setUp(self) -> None:
             {
                 "title": "aa",
                 "description": "Test Dataset",
-                "project": self.project["identifier"]
+                "project": self.project["identifier"],
+                "conditions_of_access": {},
```
davidlougheed: create a new invalid payload rather than overloading this one to have two invalid fields, since then one of them ends up getting untested just by order-of-procedure.

chord_metadata_service/chord/tests/test_api.py
```diff
@@ -167,6 +170,18 @@ def test_create_dataset_forbidden(self):
...
 
+    def test_conditions_of_access(self):
+        for d in self.valid_payloads:
+            r = self.one_authz_post("/api/datasets", json=d)
+
+            self.assertEqual(r.status_code, status.HTTP_201_CREATED)
+            self.assertEqual(Dataset.objects.get(title=d["title"]).conditions_of_access, d.get("conditions_of_access"))
+
+    def test_invalid_conditions_of_access(self):
+        d = self.invalid_payloads[0]
```
davidlougheed: avoid oblique references via index to test cases; instead I'd create a new test case variable like `self.invalid_dataset_coa` or something and then use it here directly, as well as adding it to invalid test cases

chord_metadata_service/chord/tests/test_api.py
davidlougheed: also, add a comment indicating why this is an invalid test case.

davidlougheed (changes requested):

davidlougheed (approved):
lgtm!

v-rocheleau (approved):
LGTM!

PR #594, opened 2025-09-22
Corrected Experiment extra_properties to use the correct schema

SanjeevLakhwani:


chord_metadata_service/experiments/schemas.py
```diff
@@ -2,7 +2,7 @@
 from referencing import Registry, Resource
 from .descriptions import EXPERIMENT, EXPERIMENT_RESULT, INSTRUMENT
 from chord_metadata_service.restapi.constants import MODEL_ID_PATTERN
-from chord_metadata_service.restapi.schemas import ONTOLOGY_CLASS_LIST, KEY_VALUE_OBJECT
```
davidlougheed, line 5: is `KEY_VALUE_OBJECT` used anywhere?

davidlougheed (changes requested):
add tests for non-string/string extra properties for experiments/etc.

chord_metadata_service/experiments/schemas.py
SanjeevLakhwani: it wasn't I removed it

chord_metadata_service/restapi/schema_ref.py
```diff
@@ -49,9 +48,9 @@ class SchemaRefs(Enum):
     EXPERIMENT_RESULT_FILE_INDEX_LIST_SCHEMA = EXPERIMENT_RESULT_FILE_INDEX_LIST_SCHEMA
     EXPRESSION_SCHEMA = EXPRESSION_SCHEMA
     EXTENSION_SCHEMA = EXTENSION_SCHEMA
+    KEY_VALUE_OBJECT = EXTRA_PROPERTIES_SCHEMA
```
davidlougheed: move back down to original position to keep diff + add a comment explaining why this is a mismatch; should we just put the original schema back?

davidlougheed (changes requested):

chord_metadata_service/restapi/schema_ref.py
davidlougheed: actually probably the alias is fine, but needs explaining.

davidlougheed (approved):

PR #597, opened 2025-10-09
 feat(phenopackets): make subject optional

davidlougheed:
https://redmine.c3g-app.sd4h.ca/issues/2629


gsfk (commented):
Ingestion works, it shows up in the UI correctly, ~~but some bento infrastructure is missing for data without subjects. Not sure if you want to handle in this pr or not.~~

~~In particular some of the mapping paths in discovery config seem to expect to begin with `individual`, eg for measurements, existing configs use this:~~

Update: everything in fact works great

gsfk (approved):

gsfk:
bento_web search is a bit quirky: all results are assigned to a single null individual. But I'm assuming bento_web search is low priority.

PR #602, opened 2025-10-29
feat(discovery): discovery match CSV responses

davidlougheed:
This adds a `_format` query parameter to the discovery matches endpoint, which can be used like this: `/api/discovery_matches?_format=csv` 

`_format` can be `json` or `csv`, and alternatively the HTTP `Accept` header can be set to either `application/json` or `text/csv` to get back JSON/CSV content.

Right now, the CSV format is just the legacy Katsu CSV format using the existing CSV renderers and a few new ones to round out available CSV responses, rather than something a bit nicer; that work would be done in another release/PR.

There's a bit of irrelevant code thrown in here too which fixes project/dataset ID issues in match responses + fixes one or two N+1 querying issues with the Django ORM.

It also fixes an outstanding, previously-undiscovered bug with rendering phenopackets TimeElements as strings in CSVs.


chord_metadata_service/phenopackets/utils.py
```diff
@@ -3,27 +3,40 @@
...
-    elif 'id' in onset and 'label' in onset:
-        return f"{onset['label']} {onset['id']}"
-    # age range
-    elif 'start' in onset and 'end' in onset:
-        if 'age' in onset['start'] and 'age' in onset['end']:
-            return f"{onset['start']['age']} - {onset['end']['age']}"
-    else:
-        return None
+    if "age" in time_element:
+        return time_element["age"]
```
gsfk: Is this correct? Phenopackets age & age_range have inner iso8601duration fields, eg: 
```
            "time_at_last_encounter": {
                "age": {
                    "iso8601duration": "P59.4Y"
                }
            }
```
.... or is this stripped away somewhere under many layers of Django? I tried to test this one, but age doesn't appear in the CSV.

gsfk (commented):
Works well for all four entities in the UI, except for an edge case: 

**blocking(?)**: Crashes on biosamples with no individual:
```
KeyError at /api/discovery_matches
'individual'
```

Also one small question about the age to string code.

chord_metadata_service/phenopackets/utils.py
davidlougheed: good catch, thanks

gsfk (approved):

PR #603, opened 2025-11-03
Add entity counts to Project and Dataset serializers

SanjeevLakhwani:
## Summary
- Add `counts` object to ProjectSerializer with phenopackets, individuals, biosamples, and experiments counts
- Add `counts` object to DatasetSerializer with the same entity types
- Extract `_get_entity_counts` helper function to eliminate code duplication

## Testing
Test via API:
- `GET /api/projects` - Returns projects with counts
- `GET /api/projects/<id>` - Returns specific project with counts
- `GET /api/datasets` - Returns datasets with counts
- `GET /api/datasets/<id>` - Returns specific dataset with counts


davidlougheed (changes requested):
won't work, we will discuss this later.

davidlougheed:
see https://github.com/bento-platform/katsu/blob/25eff2c1c73fbb59e8bf856a2e206bfd3acfd582/chord_metadata_service/discovery/api_views.py#L547-L548 and https://github.com/bento-platform/katsu/blob/25eff2c1c73fbb59e8bf856a2e206bfd3acfd582/chord_metadata_service/discovery/api_views.py#L402 for code you should re-use.

you must not return raw counts, as this skips censorship.

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +131,44 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+    Applies authorization and threshold-based censorship to protect privacy.
+    Uses the same pattern as discovery API views for consistency.
+
+    If request is None, returns raw counts without censorship (for internal/admin use).
+    """
+    from .api_views import QueryQuerysetsCache, discovery_queryset_entity_counts, DiscoveryQuery
+    from .censorship import censor_entity_counts
+
+    if request is None:
+        return get_entity_counts_for_scope(scope)
```
davidlougheed: q: I know this is intended to violate censorship, but I'm not sure why that would be needed?

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +131,44 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+    scope: ValidatedDiscoveryScope
+) -> EntityCountOrBoolResponse:
+    """
+    Returns censored entity counts for all discovery entities within the given scope.
+    Applies authorization and threshold-based censorship to protect privacy.
+    Uses the same pattern as discovery API views for consistency.
+
+    If request is None, returns raw counts without censorship (for internal/admin use).
+    """
+    from .api_views import QueryQuerysetsCache, discovery_queryset_entity_counts, DiscoveryQuery
```
davidlougheed: don't import inside the function

chord_metadata_service/discovery/utils.py
```diff
@@ -10,16 +12,22 @@
 )
 from chord_metadata_service.chord.data_types import KatsuDataType
 
+from .constants import DISCOVERY_ENTITIES
 from .fields_utils import normalize_field_path_true_model
 from .model_lookups import DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE, DISCOVERY_ENTITY_NAMES_TO_MODEL
 from .scope import ValidatedDiscoveryScope
+from .types import EntityCounts, EntityCountOrBoolResponse
+
+logger = get_logger(__name__)
```
davidlougheed: should use logger from `chord_metadata_service.logger`

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +155,11 @@ def jsonschema_validation(value, schema, field_name=None):
 
         return validation
 
+    def get_counts(self, obj):
+        request = self.context.get('request')
```
davidlougheed: style: use double quotes

chord_metadata_service/chord/serializers.py
```diff
@@ -177,12 +186,19 @@ class ProjectSerializer(serializers.ModelSerializer):
...
+    counts = serializers.SerializerMethodField()
+
     # noinspection PyMethodMayBeStatic
     def validate_title(self, value):
         if len(value.strip()) < 3:
             raise serializers.ValidationError("Name must be at least 3 characters")
         return value.strip()
 
+    def get_counts(self, obj):
+        request = self.context.get('request')
```
davidlougheed: code style: use double quotes

chord_metadata_service/discovery/utils.py
davidlougheed: especially since the serializer can fail to get the request (because `.get(...)` is used) and then we get a security leak as a default behaviour rather than an error, which becomes insecure by default.

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +131,44 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
     return DISCOVERY_ENTITY_NAMES_TO_MODEL[entity].get_model_scoped_queryset(scope)
+
+
+def get_entity_counts_for_scope(scope: ValidatedDiscoveryScope) -> EntityCounts:
+    """
+    Returns entity counts for all discovery entities within the given scope.
+    Uses distinct() for entities that can appear multiple times through relationships.
+    """
+    return {
+        entity: get_discovery_entity_model_scoped_queryset(entity, scope).distinct().count()
```
davidlougheed: `get_discovery_entity_model_scoped_queryset` already returns a distinct queryset.

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +131,44 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+        for entity in DISCOVERY_ENTITIES
+    }
+
+
+def get_censored_entity_counts_for_scope(
```
davidlougheed: I would prefer (if we keep this utility, I'm not sure that it's needed - it should at the very least have a use in discovery api too) if this whole function was async, and then a sync version is created just for use in serializers.

davidlougheed (changes requested):

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +127,30 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+
+
+def get_censored_counts_for_serializer(
+    request: DrfRequest, scope: ValidatedDiscoveryScope
+) -> EntityCountOrBoolResponse:
+    """
+    Sync wrapper for getting censored entity counts from serializers.
+    Wraps the async get_censored_entity_counts_async function.
+    """
+    from .api_views import get_censored_entity_counts_async
```
davidlougheed: code style: no local imports, please

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +127,30 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+
+    @async_to_sync
```
davidlougheed: I don't think this needs to be a nested function

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +127,30 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+    async def _get_counts():
+        try:
+            dt_permissions = await get_discovery_data_type_permissions(request, scope)
+            lg = logger.bind(request_id=getattr(request, 'id', None))
```
davidlougheed: double quotes consistently

chord_metadata_service/discovery/utils.py
```diff
@@ -123,3 +127,30 @@ def get_discovery_entity_model_scoped_queryset(entity: DiscoveryEntity, scope: V
...
+            return await get_censored_entity_counts_async(scope, dt_permissions, query=None, lg=lg)
+        except Exception as e:
+            logger.warning(
+                "Failed to compute entity counts for serializer, returning empty dict",
+                exc_info=e,
+                scope=scope,
+                request_id=getattr(request, 'id', None),
```
davidlougheed: double quotes consistently

chord_metadata_service/chord/tests/test_api.py
```diff
@@ -1,5 +1,7 @@
 import uuid
+from django.test import override_settings
```
davidlougheed: put from-import with the rest in alpha order

.gitignore
```diff
@@ -24,6 +24,10 @@ env/
 *.swo
 .vscode/*
 
+# assitant-cli related
```
davidlougheed: typo

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +157,21 @@ def jsonschema_validation(value, schema, field_name=None):
 
         return validation
 
+    def get_counts(self, obj):
+        request = self.context.get("request")
+
+        # Only compute counts for read operations (GET), not write operations (POST/PUT/PATCH)
+        if not request or request.method not in ('GET', 'HEAD', 'OPTIONS'):
```
davidlougheed: when would request be None?
also, double quotes consistently please

chord_metadata_service/discovery/api_views.py
```diff
@@ -323,6 +323,78 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+    lg: BoundLogger | None,
+    return_raw_counts: Literal[True],
+) -> tuple[EntityCounts, EntityCountOrBoolResponse]: ...
+
+
+async def get_censored_entity_counts_async(
+    scope: ValidatedDiscoveryScope,
+    dt_permissions: DataTypeDiscoveryPermissions,
+    query: DiscoveryQuery | None = None,
+    lg: BoundLogger | None = None,
```
davidlougheed: make logger required here

chord_metadata_service/discovery/api_views.py
```diff
@@ -323,6 +323,78 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+        return_raw_counts: If True, return tuple of (raw_counts, censored_counts) for logging
+
+    Returns:
+        If return_raw_counts=False: Dictionary mapping entities to censored counts (int) or booleans
+        If return_raw_counts=True: Tuple of (raw counts dict, censored counts dict)
+    """
+    if query is None:
+        query = DiscoveryQuery(fts=None, filters={})
+
+    if lg is None:
```
davidlougheed: remove this if and just enforce callers passing a logger

chord_metadata_service/discovery/api_views.py
```diff
@@ -323,6 +323,78 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+        query: Optional discovery query with filters/FTS (defaults to empty query)
+        lg: Optional logger instance (defaults to module logger)
+        return_raw_counts: If True, return tuple of (raw_counts, censored_counts) for logging
+
+    Returns:
+        If return_raw_counts=False: Dictionary mapping entities to censored counts (int) or booleans
+        If return_raw_counts=True: Tuple of (raw counts dict, censored counts dict)
+    """
+    if query is None:
+        query = DiscoveryQuery(fts=None, filters={})
```
davidlougheed: non-blocking: make the empty query a constant I think, like `EMPTY_DISCOVERY_QUERY`, and then below this can just be simplified to `query or EMPTY_DISCOVERY_QUERY`

davidlougheed (changes requested):

chord_metadata_service/chord/serializers.py
SanjeevLakhwani: In unittests when someone instantiates serializer

chord_metadata_service/chord/serializers.py
```diff
@@ -177,12 +199,29 @@ class ProjectSerializer(serializers.ModelSerializer):
     datasets = DatasetSerializer(read_only=True, many=True, exclude_when_nested=["project"])
     project_schemas = ProjectJsonSchemaSerializer(read_only=True, many=True)
 
+    counts = serializers.SerializerMethodField()
```
davidlougheed, line 193: do you want to add this to `always_include` here too? it should be consistent with Dataset in either case (either none in `always_include` or both)

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +157,22 @@ def jsonschema_validation(value, schema, field_name=None):
...
         return validation
 
+    def get_counts(self, obj):
+        request = self.context.get("request")
+
+        # Only compute counts for read operations (GET), not write operations (POST/PUT/PATCH)
+        if not request or request.method not in ("GET", "HEAD", "OPTIONS"):
+            logger.debug(
+                "Skipping counts computation for non-GET request",
+                dataset_id=str(obj.identifier),
```
davidlougheed: move scope creation up and add `scope_repr=repr(scope)` instead (which other endpoints do, basically we want to match other endpoints that log scope so that we can consistently search structured logs based on scope)

chord_metadata_service/chord/utils.py
```diff
@@ -0,0 +1,26 @@
+from asgiref.sync import async_to_sync
+from rest_framework.request import Request as DrfRequest
+
+from chord_metadata_service.logger import logger
+from chord_metadata_service.discovery.scope import ValidatedDiscoveryScope
+from chord_metadata_service.discovery.types import EntityCountOrBoolResponse
+from chord_metadata_service.discovery.api_views import get_censored_entity_counts_async
+from chord_metadata_service.discovery.utils import get_discovery_data_type_permissions
+
```
davidlougheed, line 9: suggestion(non-blocking): add `__all__ = ["get_censored_counts_for_serializer"]`

chord_metadata_service/chord/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+@async_to_sync
+async def get_censored_counts_for_serializer(
```
davidlougheed, line 15: take logger as argument so we can inject a logger with other bindings for nicer structured logging

chord_metadata_service/chord/serializers.py
```diff
@@ -177,12 +199,29 @@ class ProjectSerializer(serializers.ModelSerializer):
...
+        if not request or request.method not in ("GET", "HEAD", "OPTIONS"):
+            logger.debug(
+                "Skipping counts computation for non-GET request",
+                project_id=str(obj.identifier),
+                method=request.method if request else None,
+            )
+            return {}
+
+        scope = ValidatedDiscoveryScope(obj, None)
+        return get_censored_counts_for_serializer(request, scope)
```
davidlougheed: bind request method and `scope_repr` as described above to a new logger object, and then pass it here - that way any logs emitted in `get_censored_counts_for_serializer` contain all structured logging info (which is one of the cool parts of structured logging - the ability to bind more and more structured data as we drill the logger through functions for better debugging)

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +157,22 @@ def jsonschema_validation(value, schema, field_name=None):
...
+            logger.debug(
+                "Skipping counts computation for non-GET request",
+                dataset_id=str(obj.identifier),
+                method=request.method if request else None,
+            )
+            return {}
+
+        scope = ValidatedDiscoveryScope(obj.project, obj)
+        # return get_censored_counts_for_serializer(request, scope)
+        return get_censored_counts_for_serializer(request, scope)
```
davidlougheed: see below - should pass a bound logger here

chord_metadata_service/discovery/api_views.py
```diff
@@ -66,6 +66,8 @@
 
 QueryExecutionResult = tuple[QuerySet, frozenset[DiscoveryEntity]]
 
+EMPTY_DISCOVERY_QUERY = DiscoveryQuery(fts=None, filters={})
```
davidlougheed, line 69: good - is there anywhere else this is used?

davidlougheed (changes requested):

chord_metadata_service/discovery/api_views.py
```diff
@@ -323,6 +325,77 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+    dt_permissions: DataTypeDiscoveryPermissions,
+    *,
+    lg: BoundLogger,
+    query: DiscoveryQuery | None = None,
+    return_raw_counts: Literal[False] = False,
+) -> EntityCountOrBoolResponse: ...
+
+
+@overload
+async def get_censored_entity_counts_async(
```
davidlougheed: rename to `get_censored_entity_counts` since there isn't a sync version

chord_metadata_service/chord/utils.py
```diff
@@ -0,0 +1,26 @@
...
+from chord_metadata_service.discovery.api_views import get_censored_entity_counts_async
+from chord_metadata_service.discovery.utils import get_discovery_data_type_permissions
+
+
+__all__ = ["get_censored_counts_for_serializer"]
+
+
+@async_to_sync
+async def get_censored_counts_for_serializer(
+    request: DrfRequest, scope: ValidatedDiscoveryScope, logger
```
davidlougheed: type hint logger (`from structlog.stdlib import BoundLogger`)

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +157,25 @@ def jsonschema_validation(value, schema, field_name=None):
...
+        scope_repr = repr(scope)
+
+        # Create bound logger with structured context
+        lg = logger.bind(
+            dataset_id=str(obj.identifier),
+            method=request.method if request else None,
+            scope_repr=scope_repr,
+        )
+
+        # Only compute counts for read operations (GET), not write operations (POST/PUT/PATCH)
```
davidlougheed: this chunk could be moved into `get_censored_counts_for_serializer` making request type hinted as `DrfRequest | None` to avoid duplicate code

chord_metadata_service/chord/serializers.py
```diff
@@ -151,6 +157,25 @@ def jsonschema_validation(value, schema, field_name=None):
 
         return validation
 
+    def get_counts(self, obj):
+        request = self.context.get("request")
+        scope = ValidatedDiscoveryScope(obj.project, obj)
+        scope_repr = repr(scope)
+
+        # Create bound logger with structured context
+        lg = logger.bind(
```
davidlougheed: if you remove the scope-specific bindings (dataset_id/project_id since they're in scope_repr already), this could be moved into `get_censored_counts_for_serializer` alongside the repr(scope) call

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #605, opened 2025-11-12
feat(discovery): improve match results completeness + small refact

davidlougheed:
requesting code review + making sure matches still works as intended.

* adds a bunch of fields to the experiment result response from matches for use in Bento Public (basically just recreating the model in Pydantic at this point I realize; would be good to merge all of these model definitions or just use DRF serializers even though they're terrible perf-wise)
* fixes some weird edge cases with the Django ORM + fixes some related type hinting
* refactors some code added in a prev PR to re-use the query querysets cache object

### DIFF
--- a/chord_metadata_service/discovery/pydantic_models.py
+++ b/chord_metadata_service/discovery/pydantic_models.py
@@ -83,8 +83,11 @@ class MatchExperiment(BaseMatchModel):
     Compact representation of an experiment for returning/rendering search responses.
     """
     id: str = Field(..., title="Experiment ID")
-    study_type: str
+    experiment_type: str = Field(..., title="Experiment Type")
+    study_type: str = Field(..., title="Study Type")
     results: list[MatchExperimentResult]
+    # backlink:
+    phenopacket: str | None = Field(..., title="Phenopacket ID")
 
 
 class MatchBiosample(BaseMatchModel):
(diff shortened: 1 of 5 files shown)


chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -67,24 +70,50 @@ class BaseMatchModel(BaseModel, abc.ABC):
...
     """
     Compact representation of an experiment for returning/rendering search responses.
     """
     id: str = Field(..., title="Experiment ID")
-    study_type: str
+    experiment_type: str = Field(..., title="Experiment Type")
+    study_type: str = Field(..., title="Study Type")
     results: list[MatchExperimentResult]
+    # backlink:
+    phenopacket: str | None = Field(..., title="Phenopacket ID")
```
SanjeevLakhwani, line 116: For the fields that have `None` as a possible type are they supposed to be required as indicated by the `...` in the Field object

gsfk (approved):
Behaving well in testing, tried very hard to confuse or crash it and got nowhere.

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: this is intended

PR #606, opened 2025-11-13
Features/discovery return counts by ds

noctillion:
Here the discovery endpoint was extended to support per-dataset entity counts, improving granularity for multi-dataset environments. It introduces helpers for fetching and censoring by-dataset counts, and adds them to the response

### DIFF
--- a/chord_metadata_service/discovery/api_views.py
+++ b/chord_metadata_service/discovery/api_views.py
@@ -30,7 +30,7 @@ from chord_metadata_service.restapi.pagination import DEFAULT_PAGE_SIZE, DEFAULT
 from chord_metadata_service.utils import build_id_set
 
 from . import responses as dres
-from .censorship import get_rules, thresholded_count, censor_entity_counts
+from .censorship import get_rules, thresholded_count, censor_entity_counts, censor_entity_counts_by_dataset, aggregate_counts_from_censored_by_dataset
 from .constants import DISCOVERY_ENTITIES, ENTITY_TO_DATASET_GROUP_BY
 from .exceptions import DiscoveryScopeException
 from .fields import get_field_options, get_range_stats, get_categorical_stats, get_date_stats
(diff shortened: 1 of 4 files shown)


chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -150,6 +150,7 @@ class DiscoveryResponse(BaseModel):
         dict[str, EntityCountOrBoolResponse] |
         dict[str, dict[str, EntityCountOrBoolResponse]]
```
davidlougheed, line 151: this typing was already meant to be by project and dataset, so perhaps you could re-use this type hint. long term, we probably only ever want to return counts by the most nested the user has access to (by dataset with dataset access, etc.)

chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -150,6 +150,7 @@ class DiscoveryResponse(BaseModel):
...
     )
+    counts_by_dataset: Optional[Dict[str, Dict[DiscoveryEntity, int | bool]]] = Field(default_factory=dict)
```
davidlougheed: keep in mind right now our permissions system lets users only have project-level counts, not dataset-level counts. we need to decide how to handle this - maybe by doing by project, by dataset, etc. all as separate fields which are None if they don't have permissions. not sure yet.

chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -3,7 +3,7 @@
 from bento_lib.discovery import FieldDefinition, OverviewSection, DiscoveryEntity, SearchSection
 from pydantic import BaseModel, Field, RootModel
 from rest_framework.request import Request as DrfRequest
-from typing import TypeAlias
+from typing import TypeAlias, Optional, Dict
```
davidlougheed: this is old-fashioned type hinting (Dict from typing and Optional) - just use `dict[...]` and union types (`a | b`)

chord_metadata_service/discovery/api_views.py
```diff
@@ -350,6 +350,59 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+        all_datasets.update(ec.keys())
+
+    res = {}
+    for ds in all_datasets:
+        res[ds] = {list(DISCOVERY_ENTITIES)[i]: entity_counts[i].get(ds, 0) for i in range(len(DISCOVERY_ENTITIES))}
+
+    return res
+
+
+async def censor_entity_counts_by_dataset(
```
davidlougheed: there should be already a lot of code which handles counts censorship, which should be reused

chord_metadata_service/discovery/api_views.py
```diff
@@ -350,6 +350,59 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
     }
 
 
+async def discovery_queryset_entity_counts_by_dataset(
```
davidlougheed, line 359: this could be done with existing code for entity counting I believe

davidlougheed (changes requested):

davidlougheed (changes requested):
this code doesn't currently handle the case where users do not have dataset level counts permissions... we need to consider perhaps returning a nested structure with projects and datasets.

chord_metadata_service/discovery/api_views.py
```diff
@@ -350,6 +356,40 @@ async def _get_entity_count(ee: DiscoveryEntity) -> int:
...
+    entity_counts_per_entity = await asyncio.gather(
+        *(_get_entity_counts_by_dataset(e) for e in DISCOVERY_ENTITIES)
+    )
+
+    all_datasets: set[str] = set()
+    for ec in entity_counts_per_entity:
+        all_datasets.update(ec.keys())
+
+    res: dict[str, EntityCounts] = {}
+    entities = tuple(DISCOVERY_ENTITIES)
```
davidlougheed: this isn't needed, you can use `DISCOVERY_ENTITIES` directly

noctillion:
comments adressed an changes moved to features/discovery_return_counts_by_dataset pr 666

PR #607, opened 2025-11-14
feat(discovery): faster + better full-text search; discovery refact

davidlougheed:
* The large diff here doesn't include any changed code to be reviewed, I just shuffled code around for organization's sake and to prevent circular imports --> ignore test diffs + stuff in `chord_metadata_service/discovery/field_paths/*`
* FTS queries with the demo dataset should take maybe 3 seconds at most.
* The first time this is merged/updated-to, one **must** run `./manage.py populate_fts` in the Katsu shell. This will need to be put in the migration guide too.
* This adds the discovery API `_fts_type` query parameter for running more advanced full-text search
  * See https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/search/#searchquery and https://www.postgresql.org/docs/18/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES


gsfk (approved):
Works great!

For the demo dataset, lots of searches now take in the 1s range. This is a real improvement on current free text search (consistently around 11 or 12s for demo dataset on staging).

Performed well in all the cases I tried: 
- different ontology term fields
- extra properties
- experiment & experiment results fields 
- random other bits, including field names

Most searches on demo dataset took less than 1s. The slowest results were for results that returned everything, eg fts=Homo+sapiens, those took about 2.5s.

I found a few fields that are not covered by search, such as the interpretation subfields `gene_descriptor` and `variant_interpretation`, but this looks at most like a future TODO.

The big test will be on data with a large extra properties field, such as BQC. Free text search on current BQC is in the order of 3 minutes.

chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -442,6 +447,9 @@ def ingest_phenopacket(
     phenopacket.biosamples.set(biosamples_db)
     phenopacket.diseases.set(diseases_db)
 
+    # ... and re-save it, to populate the fts_extra field from the above related models
+    phenopacket.save()
+
```
gsfk, line 452: is there a lore reason why save() here is from the BaseExtraProperties class?

chord_metadata_service/chord/management/commands/populate_fts.py
```diff
@@ -0,0 +1,29 @@
...
+        Populates fts_extra fields for all discovery entity models.
+    """
+
+    def handle(self, *args, **options):
+        for e, m in DISCOVERY_ENTITY_NAMES_TO_MODEL.items():
+            self.stdout.write(f"Working on entity {e}\n")
+
+            qs = m.objects.all()
+            if e == "biosample":
+                qs = qs.prefetch_related("phenotypic_features").select_related("location_collected")
```
gsfk, line 18: I don't understand why `phenotypic_features` and `location_collected` are together here

chord_metadata_service/chord/management/commands/populate_fts.py
```diff
@@ -0,0 +1,29 @@
...
+            elif e == "phenopacket":
+                qs = (
+                    qs
+                    .prefetch_related("interpretations", "diseases", "phenotypic_features")
+                    .select_related("meta_data")
+                )
+            elif e == "experiment":
+                qs = qs.select_related("instrument")
+            elif e == "individual":
+                qs = qs.select_related("vital_status")
```
gsfk, line 28: Some fields get singled out here, but i's not clear to me why they get special handling

chord_metadata_service/discovery/field_paths/django_field_query.py
```diff
@@ -0,0 +1,93 @@
...
+def get_field_django_mapping_and_queried_entity(
+    queryset_entity: DiscoveryEntity, field_props: AnyFieldDefinition, force_through_phenopackets: bool = False
+) -> tuple[str, DiscoveryFieldSubquery | None, DiscoveryEntity]:
+    """
+    Parses a path-like string representing an ORM such as "individual/extra_properties/date_of_consent"
+    where the first crumb represents the object in the DB model, and the next ones
+    are the field with their possible joins through tables relations.
+    Returns a tuple of (
+        the Django string representation of the field for this object relative to the queryset entity,
+        a specification for executing an Exists(...) subqueyr IF crossing a many-to-many or many-to-one boundary,
```
gsfk: "subqueyr"

chord_metadata_service/discovery/field_paths/resolve.py
```diff
@@ -0,0 +1,237 @@
...
+            }:
+                return field_path[2:]
+            return "experiments", "biosample", *field_path
+        case ("experiment_result", "experiment"):
+            if field_path[:1] == ("experiment_results",):
+                return field_path[1:]
+            return "experiments", *field_path
+        # --------------------------------------------------------------------------------------------------------------
+        case _:
+            raise exc
```
gsfk, line 188: General question about the conditions found in many of the cases here, of the form`if field_path[:n] in`, eg from `case ("experiment_result", "biosample")`
```
            if field_path[:4] in {
                ("phenopackets", "biosamples", "experiment", "experiment_results"),  # TODO: remove in future version
                ("phenopackets", "biosamples", "experiments", "experiment_results"),
            }:
```

When do these cases actually happen? I tried different queries with different configs and could never get anything like them. In particular for the case above (queryset_entity: experiment_result, field_entity: biosample) the field_path always begins with field internal to biosample:

`field_path: ('sampled_tissue', 'label')`

and could never get anything like what this condition expects. Or would the field path be longer if the mapping in my config file were longer?

chord_metadata_service/phenopackets/tests/constants.py
```diff
@@ -498,16 +498,16 @@ def valid_phenotypic_feature(biosample=None, phenopacket=None):
...
         modifiers=[
             {
-                "id": "HP: 0012825 ",
+                "id": "HP:0012825",
                 "label": "Mild"
             },
             {
-                "id": "HP: 0012823 ",
+                "id": "HP:0012823",
                 "label": "Semi-mild"
```
gsfk: old code and probably not relevant to the test, but `HP:0012823` is the ontology for a modifier, rather than a modifier itself. I'm guessing this is meant to be `HP:0012826` - "Moderate" or something similar.

chord_metadata_service/phenopackets/models.py
```diff
@@ -370,6 +444,9 @@ class GenomicInterpretation(BaseTimeStamp):
     extra_properties = JSONField(blank=True, null=True,
                                  help_text='Extra properties that are not supported by current schema')
 
+    def fts_repr_values(self) -> tuple:
+        return self.interpretation_status, self.gene_descriptor, self.variant_interpretation, self.extra_properties
+
```
gsfk, line 454: Code looks fine here, but genomic_interpretation fields are missing from free text search, as are all the fields from variant_interpretation. Possibly a problem with how fts is populated?

gsfk (commented):
Some code comments as requested. There's a lot of katsu background I'm missing, so i just picked out everything I didn't understand.

chord_metadata_service/discovery/field_paths/resolve.py
gsfk: ... the experiment / experiments thing I understand, just not clear to me where these long paths are expected to come from

davidlougheed: to be honest, this will never occur in a config written normally because no one is going to put `biosample/phenopackets/biosamples/...` - this was for the sake of completeness but basically just dumb overkill.

chord_metadata_service/chord/management/commands/populate_fts.py
davidlougheed: i'll add a comment; prefetches/select related are for django performance with joined models

chord_metadata_service/chord/management/commands/populate_fts.py
davidlougheed: foreign keys need `select_related` to avoid the n+1 query ORM thing

chord_metadata_service/discovery/field_paths/django_field_query.py
davidlougheed: fixing, thanks

chord_metadata_service/phenopackets/models.py
davidlougheed: there's supposed to be a recursive fts_extra population mechanism, but maybe not working here...

chord_metadata_service/phenopackets/tests/constants.py
davidlougheed: fixing

chord_metadata_service/chord/management/commands/populate_fts.py
davidlougheed: added a comment

chord_metadata_service/chord/ingest/phenopackets.py
davidlougheed: I need to look into how multiple inheritance works here...

davidlougheed: oops

davidlougheed: should be fixed now...

chord_metadata_service/phenopackets/models.py
davidlougheed: is there an example you can give me about how this doesn't get populated properly?

chord_metadata_service/chord/management/commands/populate_fts.py
gsfk: but isn't location_collected strictly a biosample property? What's the connection with phenotypic_features?

davidlougheed: they're not connected - `prefetch_related` and `select_related` return modified copies of the queryset the functions were called on, so this essentially just chains modifications to the queryset together.

`qs.prefetch_related("phenotypic_features")` returns the Biosample `qs` but with that prefetch, and then the subsequent `select_related` returns the modified Biosample qs now with both the prefetch and the `select_related`

chord_metadata_service/phenopackets/models.py
gsfk: a bit broad to qualify as an example maybe, but for demo dataset I get zero results for strings that are only found in gene_descriptor, such as "HGNC:2867", "DHODH":


example dataset field:
```
                            "gene_descriptor": {
                                "value_id": "HGNC:2867",
                                "symbol": "DHODH"
                            }
```

... and similarly for variant_descriptor fields. But I can't tell if the issue is whether it's populated properly or not. Is there a way I can tell in adminer?

davidlougheed: take a look at the `fts_extra` field on phenopackets and see if it's populated. did you run the command in the container as described in the pr desc.?

gsfk: pulled and reran `populate_fts` and still zero results for "DHODH"

davidlougheed: what is the value for the phenopackets `fts_extra` field?

gsfk: it's a bit squirrely: 

FAKE-049-interpretation:001 SOLVED FAKE-049-interpretation:001 SNOMED:66038001 Miller syndrome <QuerySet [<GenomicInterpretation: 1>]> SNOMED:66038001 Miller syndrome SNOMED:840539006 COVID-19 SNOMED:422587007 Nausea C3G_synthetic_data

davidlougheed: ah I see the issue though - that stringified queryset

davidlougheed: thanks!

gsfk: working now

chord_metadata_service/chord/management/commands/populate_fts.py
gsfk: righto

gsfk (approved):

PR #608, opened 2025-11-18
refact(discovery): QueryHelper with counts utils + nested field censoring

davidlougheed:
* renames QueryQuerysetsCache to QueryHelper
* encapsulates some entity count methods in QueryHelper
* censors nested entity/field results if parent entity is censored (e.g., biosamples are *now* always censored if individuals are censored)

### DIFF
--- a/chord_metadata_service/discovery/api_views.py
+++ b/chord_metadata_service/discovery/api_views.py
@@ -71,7 +71,7 @@ is_not_none = partial(is_not, None)
 
 QueryExecutionResult = tuple[QuerySet, frozenset[DiscoveryEntity]]
 
-EMPTY_DISCOVERY_QUERY = DiscoveryQuery(fts=None, filters={})
+EMPTY_DISCOVERY_QUERY = DiscoveryQuery()
 
 
 class QueryHelper:
@@ -152,7 +152,7 @@ class QueryHelper:
 
         fts_queried_entities = frozenset()
 
-        if fts := self._query.fts:
+        if fts := self._query.fts:  # Execute FTS if self._query.fts is not ""
             # Each entity has a corresponding FTS vector we'll use, but querying using this doesn't query across
             # discovery entity boundaries. In order to get result sets for, e.g., phenopackets (with biosamples being
             # the entity with actual FTS matches), we need to build a query for the queryset checking nested biosamples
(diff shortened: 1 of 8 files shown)


chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -216,6 +216,9 @@ class DiscoveryQuery(BaseModel):
     def queried_filter_fields(self) -> list[str]:
         return list(self.filters.keys())
 
+    def is_empty(self) -> bool:
+        return self.fts is None and len(self.filters) == 0
```
v-rocheleau: **question:** does this handle cases where `self.fts == ""` correctly? If `fts` is an empty string and we have no `filters` this will return `False`.

chord_metadata_service/discovery/api_views.py
```diff
@@ -75,21 +74,26 @@
 EMPTY_DISCOVERY_QUERY = DiscoveryQuery(fts=None, filters={})
 
 
-class QueryQuerysetsCache:
```
v-rocheleau, line 78: Good rename lol

chord_metadata_service/discovery/api_views.py
```diff
@@ -193,6 +215,100 @@ async def get_query_queryset_and_queried_entities(
...
+
+    @overload
+    async def get_censored_entity_counts(self, return_raw_counts: Literal[False] = False) -> EntityCountOrBoolResponse:
+        ...
+
+    @overload
+    async def get_censored_entity_counts(
+        self, return_raw_counts: Literal[True]
+    ) -> tuple[EntityCounts, EntityCountOrBoolResponse]:
+        ...
```
v-rocheleau, line 256: 🤯

v-rocheleau (commented):
Quick pass, looks good but I'll take a closer look tomorrow

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: good point!

chord_metadata_service/discovery/api_views.py
davidlougheed: 💟

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: now I just use blank string as the empty value instead of having `str | None`

v-rocheleau (approved):
LGTM!

PR #609, opened 2025-11-27
feat(discovery): more linking across entities in match responses

davidlougheed:
adds experiment + phenopacket backlinks from experiment results in the discovery matches API.


v-rocheleau (approved):
LGTM!

davidlougheed:
thanks!

PR #611, opened 2025-11-28
fix(cleanup): add missing cleanup for variant interpretations

davidlougheed:
variant interpretations weren't getting cleaned up, which was causing ingestion errors if it changed subtly.


v-rocheleau (approved):
LGTM!

PR #616, opened 2025-12-04
feat(discovery): support "one or more of"-type queries

davidlougheed:


gsfk (approved):

gsfk (commented):
A few small quirks remaining, importance of each is left to the reader: 

- If you fill in a few filters and then go back and start removing tags, it's now possible to have filters shown in the UI with search fields ("Sex", "Disease", etc) but no values. This is probably fine but when removing a filter (any filter at all), if any of your filters have no values, those are removed as well.

- Changing the field on an existing filter reorders it to the bottom (eg if I change "Sex" to "Disease" instead of deleting Sex and adding Disease) (hmm... checking again it seems like this is already an issue in prod version)

- adding lots of tags to one filter seems to slow katsu down more than adding lots of filters, and the UI tends to hang while waiting (I can't add more filters while waiting, for example). The waits are not wildly long, however.

chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -221,6 +224,20 @@ class DiscoverySearchFieldsResponse(BaseModel):
...
+    model_config = ConfigDict(frozen=True)
+
+    filter_type: Literal["one_of"]
+    negated: bool = False
+
+
+class DiscoveryQueryFilterOneOf(DiscoveryQueryFilterBase):
+    model_config = ConfigDict(frozen=True)
+
+    filter_type: Literal["one_of"]  # really more like "one or more of" - essentially Boolean Or for filter values
```
v-rocheleau, line 237: **suggestion:** both paramethers are set the same and inherited from `DiscoveryQueryFilterBase`. Though its probably best to define the the `filter_type` literal here

chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -221,6 +224,20 @@ class DiscoverySearchFieldsResponse(BaseModel):
     sections: list[DiscoverySearchSectionWithOptions]
 
 
+class DiscoveryQueryFilterBase(BaseModel):
+    model_config = ConfigDict(frozen=True)
+
+    filter_type: Literal["one_of"]
+    negated: bool = False
```
v-rocheleau, line 231: Nice, great to have the negation on top!
Does it work with two query filters, a "one-ofs" and a "not-one-ofs" at the same time?

chord_metadata_service/discovery/pydantic_models.py
```diff
@@ -236,25 +253,51 @@ class DiscoveryQuery(BaseModel):
...
+    def n_filter_parameters(self) -> int:
+        """
+        Returns the number of filter parameters in the query; more than one query to the same field in an OR fashion
+        counts as multiple parameters.
+        """
+        n = 0
+        for f in self.filters.values():
+            if isinstance(f, DiscoveryQueryFilterOneOf):
+                n = sys.maxsize  # cannot be used in a censored discovery context
+                break
```
v-rocheleau, line 277: **question:** sorry not sure I get this part, think I'm just rusty on how censorship works here.

Why set the number of filter params to `maxsize` if any of the filters values in a query is a `DiscoveryQueryFilterOneOf`, and how does that relate to censorship?

v-rocheleau (commented):
Quick first pass before ending the day but looks good!

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: this doesn't actually end up getting used, so I think you're right that we should maybe make this clearer. the problem with one-of searches is that it lets us do subtraction to figure out low cell counts (`A = 5, B = censored; A | B is 6 --> B is 1`). this is checked beforehand, so this branch doesn't actually get hit in the code. I'll try to see if there's a nicer way of making it clear this is already enforced.

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: it would only negate the one-of filter, so `A = ![one of (a, b, c)], B = 'hello'` - would give all records matching `B = 'hello'` that do not have any `A` in `(a, b, c)`. Negation support is actually a future ticket, but I put this here to build off of at that point. I may still add some more tests for this though.

chord_metadata_service/discovery/pydantic_models.py
davidlougheed: this is more for future-proofing, when we add `AllOf` queries (AND rather than OR). Then, the base class will have a broader `filter_type` than the derived classes, in essence.

v-rocheleau (approved):
LGTM!

PR #619, opened 2025-12-09
fix(ingest): fix genomic interpretations bad reuse of existing objects

davidlougheed:
fixes genomic interpretation objects being re-used more often than they should have.


gsfk:
🔍

gsfk (approved):
Looks good. 

Data will have to be re-ingested on any instance with interpretation data, although I can't think of a prod instance that actually has interpretations.

PR #625, opened 2026-01-07
fix(patients): properly serialize vital status

davidlougheed:
see https://redmine.c3g-app.sd4h.ca/issues/2692


gsfk (approved):
works great. Bento public rendering of vital status needs some improvement.

PR #632, opened 2026-01-14
chore: sort discovery field options when enum is null

davidlougheed:


gsfk (changes requested):
can we do case-insensitive sort?

gsfk (approved):

PR #635, opened 2026-01-14
feat!: new experiments model aligned with GA4GH

noctillion:
This PR aligns the experiment model with GA4GH standards by introducing new metadata fields such as library_id, insert_size, and protocol_url. It refactors molecule_ontology and experiment_ontology from lists to single objects, including a consolidated database migration with a script to convert legacy data. Te ingestion ans tests were modified to support changes.


chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -89,18 +90,25 @@ def ingest_experiment(
     new_experiment_id = experiment_data.get("id", str(uuid.uuid4()))
     study_type = experiment_data.get("study_type")
     experiment_type = experiment_data["experiment_type"]
-    experiment_ontology = experiment_data.get("experiment_ontology", [])
+    experiment_ontology = experiment_data.get("experiment_ontology", {})
```
davidlougheed: 1. the default should not be an empty dictionary, but None instead (which is the default for `.get`), since an empty dictionary is not a valid ontology term.
2. can you add a fallback if this is an array, to just grab the first element? that way we can preserve backwards-compatibility with old experiments. if the array is empty, set this to None.

chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -89,18 +90,25 @@ def ingest_experiment(
...
     molecule = experiment_data.get("molecule")
-    molecule_ontology = experiment_data.get("molecule_ontology", [])
+    molecule_ontology = experiment_data.get("molecule_ontology", {})
```
davidlougheed: idem

chord_metadata_service/experiments/descriptions.py
```diff
@@ -15,14 +17,14 @@
                            "Histone H3K4me1, etc.).",
         "experiment_ontology": {
```
davidlougheed: this can be simplified, since the argument to `ontology_class` is already the description prefixed with "An ontology class" basically

so this would be just

```
"experiment_ontology": ontology_class("describing the experiment, e.g., via the OBI ontology")
```

chord_metadata_service/experiments/descriptions.py
```diff
@@ -15,14 +17,14 @@
...
-            "items": ontology_class("describing the experiment"),
+            **ontology_class("describing the experiment"),
         },
         "molecule": "(Controlled Vocabulary) The type of molecule that was extracted from the biological material."
                     "Include one of the following: total RNA, polyA RNA, cytoplasmic RNA, nuclear RNA, small RNA, "
                     "genomic DNA, protein, or other.",
         "molecule_ontology": {
             "description": "Links to molecule ontology information (e.g. via the SO ontology.).",
-            "items": ontology_class("describing a molecular property"),
+            **ontology_class("describing a molecular property"),
```
davidlougheed:
```
"molecule_ontology": ontology_class("describing the molecule, e.g., via the SO ontology")
```

chord_metadata_service/experiments/migrations/0014_v13_0_0.py
```diff
@@ -0,0 +1,109 @@
...
+    Iterates through all experiments. If ontology fields are lists,
+    converts them to a single object (the first item) or an empty dict.
+    """
+    # retrieve the historical version of the model
+    Experiment = apps.get_model('experiments', 'Experiment')
+    
+    for exp in Experiment.objects.all():
+        # fix molecule_ontology
+        mol_data = exp.molecule_ontology
+        if isinstance(mol_data, list):
```
davidlougheed, line 20: nice!

chord_metadata_service/experiments/models.py
```diff
@@ -60,8 +59,8 @@ def get_scope_filters() -> ModelScopeFilters:
     )
     experiment_ontology = JSONField(
         blank=True,
-        default=list,
-        validators=[ontology_list_validator],
+        default=dict,
```
davidlougheed: default should be None, and this should be made nullable

chord_metadata_service/experiments/models.py
```diff
@@ -73,8 +72,8 @@ def get_scope_filters() -> ModelScopeFilters:
     )
     molecule_ontology = JSONField(
         blank=True,
-        default=list,
-        validators=[ontology_list_validator],
+        default=dict,
```
davidlougheed: idem

chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -132,10 +140,17 @@ def ingest_experiment(
...
+        library_extract_id=library_extract_id,
+        insert_size=insert_size,
+        description=description,
+        protocol_url=protocol_url,
+        library_description=library_description,
         extraction_protocol=extraction_protocol,
         reference_registry_id=reference_registry_id,
         qc_flags=qc_flags,
         biosample=biosample,
+        ga4gh_template_id=ga4gh_template_id,
```
davidlougheed: i would remove this field, as the schema is dictated by our database structure and so it doesn't make sense to have it user-settable in my opinion, unless I'm missing something/.

chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -89,18 +90,25 @@ def ingest_experiment(
...
     library_strategy = experiment_data.get("library_strategy")
     library_source = experiment_data.get("library_source")
     library_selection = experiment_data.get("library_selection")
     library_layout = experiment_data.get("library_layout")
+    library_id = experiment_data.get("library_id")
+    library_extract_id = experiment_data.get("library_extract_id")
+    insert_size = experiment_data.get("insert_size")
+    description = experiment_data.get("description")
+    protocol_url = experiment_data.get("protocol_url")
+    library_description = experiment_data.get("library_description")
```
davidlougheed, line 104: can you sort these fields a bit, maybe adding a comment subheading like the following to group all library fields under, and then keep relations together (biosample/experiment results/instrument), etc? this function is a bit dense and could use some commenting/whitespace structure to improve readability.

```
# library fields
...
```

davidlougheed (changes requested):
looking good so far! see specific code comments, plus:

* add new fields to `FULL_TEXT_SEARCH_FIELDS` in `discovery.full_text_search`, casting any ontology terms to text fields
  * don't add `insert_size` as an fts field since it isn't very useful to be able to free-text search on it.

we will also need to update the demo dataset to match this at some point.

chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -89,16 +90,25 @@ def ingest_experiment(
...
     library_strategy = experiment_data.get("library_strategy")
     library_source = experiment_data.get("library_source")
     library_selection = experiment_data.get("library_selection")
     library_layout = experiment_data.get("library_layout")
+    library_id = experiment_data.get("library_id")
+    library_extract_id = experiment_data.get("library_extract_id")
+    insert_size = experiment_data.get("insert_size")
+    library_description = experiment_data.get("library_description")
+    # protocol fields
+    description = experiment_data.get("description")
```
davidlougheed: nitpick: move this line up right below experiment ID i think? since it's more general and not part of protocol

chord_metadata_service/chord/ingest/experiments.py
```diff
@@ -132,6 +142,12 @@ def ingest_experiment(
         library_source=library_source,
         library_selection=library_selection,
         library_layout=library_layout,
+        library_id=library_id,
+        library_extract_id=library_extract_id,
+        insert_size=insert_size,
+        description=description,
```
davidlougheed: idem

chord_metadata_service/experiments/schemas.py
```diff
@@ -177,6 +177,30 @@
...
+        "description": {
+            "type": "string",
+            "description": "Experimental design description."
+        },
+        "protocol_url": {
+            "type": "string",
+            "format": "uri",
+            "description": "URL to the protocol."
+        },
+        "library_description": {
```
davidlougheed: move up with other `library_` fields

chord_metadata_service/experiments/search_schemas.py
```diff
@@ -131,6 +131,24 @@
...
+        "description": {
+            "search": search_optional_str(order=17),
+        },
+        "protocol_url": {
+            "search": search_optional_str(order=18),
+        },
+        "library_description": {
+            "search": search_optional_str(order=19),
+        },
+        "library_extract_id": {
```
davidlougheed: idem for these two, sorry

chord_metadata_service/experiments/descriptions.py
```diff
@@ -33,6 +28,12 @@
         "library_selection": "Method used to enrich the target in the sequence library preparation. "
                              "E.g. Random, PCR, Random PCR, RT-PCR, MF and other.",
         "library_layout": "The library layout. E.g. Single, Paired.",
+        "library_id": "The identifier associated with the sequencing library.",
+        "library_extract_id": "Optional identifier for a given specific extraction.",
+        "insert_size": "The size of the DNA fragment inserted into the vector (Integer).",
+        "description": "Free-text experimental design description.",
```
davidlougheed: move up below ID

chord_metadata_service/experiments/descriptions.py
```diff
@@ -33,6 +28,12 @@
...
+        "protocol_url": "URL to the sequencing protocol.",
+        "library_description": "Specific free-text details about library construction.",
```
davidlougheed: move with other library fields

chord_metadata_service/chord/ingest/utils.py
```diff
@@ -37,6 +38,13 @@ def map_if_list(fn: Callable, data: Any, *args, **kwargs) -> Any:
     )
 
 
+def get_ontology_term(data, key):
```
davidlougheed: add type hint to `key: str`
also, rename this since it's slightly more specific than just getting an ontology term - perhaps `get_single_ontology_term`? then, add a comment above the if statement explaining that this is for the previous experiment schema, where some fields were lists of single ontology items. otherwise, the existence of this function in this file will leave future devs a bit confused - i think some historical context in a comment is needed here.

davidlougheed (changes requested):

davidlougheed (approved):
lgtm!

PR #636, opened 2026-01-15
fix: correct updates schema + ingest phenopacket metadata updates

davidlougheed:
see https://phenopacket-schema.readthedocs.io/en/latest/update.html#rstupdate

### DIFF
--- a/chord_metadata_service/chord/ingest/phenopackets.py
+++ b/chord_metadata_service/chord/ingest/phenopackets.py
@@ -436,6 +436,7 @@ def ingest_phenopacket(
         submitted_by=meta_data.get("submitted_by"),
         phenopacket_schema_version=meta_data.get("phenopacket_schema_version"),
         external_references=meta_data.get("external_references", []),
+        updates=meta_data.get("updates", []),
         extra_properties=remove_computed_properties(meta_data.get("extra_properties", {})),
     )
     meta_data_obj.save()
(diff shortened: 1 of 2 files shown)


gsfk (commented):
Working as expected, so no problem there. 

But inspecting the raw katsu response to see if it's working leads to a couple questions:

- The metadata field in the response already includes a field called "updated" which is not in the phenopacket schema. So now we have this situation: 

  ```

  	"meta_data": {
		  "updated": <some timestamp, typically ingestion date>,
		  "updates": [
			{
				"timestamp": <some unrelated timestamp>
			}
		]
    ```
    do we still need the old "updated" field? It's not clear to me why it's there. Best guess is that it's for PUT / PATCH calls that we're not using anyway.

- Looking over the entire phenopacket response there are a _huge_ number of "created" and "updated" fields scattered throughout the phenopacket, and again it's not clear that they are needed (they're in every biosample, every phenotypic feature, every metadata resource... ). Do we need these in the response? I was able to remove an entire kB of data from a 4kB phenopacket just by deleting all of these.

davidlougheed:
that's an old remnant yes, and tied to put/patch/whatever. I agree we maybe don't need them, although that's separate from this PR. we should maybe talk about it next week? it's on pretty much every db model, so I don't really want to pull it out here.

gsfk:
Yeah, I think we can merge this now and deal with the rest later.

PR #651, opened 2026-02-04
feat(chord): update resources with the same ID with new ingestions

davidlougheed:
This doesn't deal with mismatched versions of resources, but it does let us change the URL of resources with the same ID.

### DIFF
--- a/chord_metadata_service/chord/ingest/resources.py
+++ b/chord_metadata_service/chord/ingest/resources.py
@@ -48,7 +48,6 @@ def ingest_resource(resource: dict, logger: BoundLogger) -> rm.Resource:
         version=version,
         iri_prefix=resource["iri_prefix"],
         extra_properties=resource.get("extra_properties", {}),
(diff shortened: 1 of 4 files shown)


chord_metadata_service/chord/ingest/resources.py
```diff
@@ -1,21 +1,53 @@
...
     rs_obj, _ = rm.Resource.objects.get_or_create(
         id=assigned_resource_id,
         name=resource["name"],
         namespace_prefix=namespace_prefix,
         url=resource["url"],
         version=version,
         iri_prefix=resource["iri_prefix"],
-        extra_properties=resource.get("extra_properties", {})
+        extra_properties=resource.get("extra_properties", {}),
         # TODO extra_properties
```
gsfk: is this TODO still to do? it looks done

gsfk (approved):
great!, with one annoying inline comment

PR #654, opened 2026-02-10
fix(discovery): handle custom decimal bins in discovery number field

davidlougheed:
auto-generated decimal bins is a more complex issue, not tackling that here.

https://redmine.c3g-app.sd4h.ca/issues/2723


gsfk (approved):
Nice!

PR #657, opened 2026-02-12
feat(chord): add experiment result/DRS object association workflow

davidlougheed:
associates experiment results with DRS objects based on filename, as an "analysis" workflow. in the future it's better to ingest experiments with files at the same time, but this helps us repair existing instances (like RENATA).


gsfk:
Ran and it returned error 400 :(

It's not immediately clear to me how the recommended approach for the future will work for vcfs, which for me at least are typically done in batches. Will I have to make a custom json for each batch?

gsfk (approved):

PR #659, opened 2026-02-20
feat(discovery): support trigram-based full-text search

davidlougheed:
ticket: https://redmine.c3g-app.sd4h.ca/issues/2752


chord_metadata_service/discovery/full_text_search.py
```diff
@@ -1,21 +1,26 @@
...
+    "entity_search_fields",
     "full_text_search_vector",
+    "normal_full_text_search",
+    "trigram_similarity_search",
     "BaseFTSModel",
     "FTSHelpersMixin",
     "ToFTSReprMixin",
 ]
 
+TRIGRAM_MINIMUM_SIMILARITY = 0.5
```
noctillion: it could be more permisive a bit.. 0.3?

davidlougheed (commented):
right now we're getting duplicate records somewhere (visible in histograms) - needs fixing
EDIT: should be fixed now

chord_metadata_service/discovery/full_text_search.py
davidlougheed: I think we can make it more permissive when we have rank in the full text search, so that we can properly show relevance of rows. does this sound good?

noctillion: Completelly, I was thinking  that 0.3 could work better

noctillion (approved):
I tested with compex substrings and works well!  Very nice implementation of the threshold adapting on query length. LGTM!

gsfk (approved):
Nice! 

A few possibly unexpected features: 

In the bento dataset I can search by age using the iso string, eg "P20Y".

For some fields it behaves almost like a Google search, "male alive" returns all males with vital status "alive". This only works for some fields, though ("male hashimoto" doesn't work)

PR #660, opened 2026-02-23
replace TrigramSimilarity with TrigramWordSimilarity

noctillion:
changes reated to feat/discovery/fts-trigram


davidlougheed (approved):
great, thanks for testing this + figuring out a better similarity function!

PR #665, opened 2026-03-08
Dataset provenance model update

SanjeevLakhwani:
## Summary

- **New `DatasetV2` model** backed by a `PydanticJSONBMixin` that splits a `ProjectScopedDatasetModel` payload between real DB columns (`identifier`, `project`, `title`, `release_date`, `last_modified`) and a validated JSONB blob (`data`). Replaces the earlier composite-PK design with a single `identifier` primary key.
- **`DatasetV2Translation` model** for per-language translations of the dataset payload, with a `(dataset, language)` unique constraint and its own CRUD API (`/datasets_v2/<id>/translations[/<lang>]`).
- **`DatasetV2ScopeAdapter`** duck-type wrapper so `DatasetV2` satisfies the `ValidatedDiscoveryScope` dataset interface without adding a `discovery` attribute to the model.
- **Full CRUD REST API** for `DatasetV2` and translations; datasets are nested under `ProjectSerializer` as `datasets_v2`.
- **`/datasets_v2/<id>/data-types`** and **`/datasets_v2/<id>/data-types/<data_type>`** endpoints mirroring the legacy dataset equivalents, including `DELETE` with authz and cleanup.
- **`/datasets_v2/<id>/summary`** discovery summary endpoint.
- **`_counts`** field on `DatasetV2Serializer` — auto-computed censored entity counts (Phenopackets/Experiments) using the same `get_censored_counts_for_serializer` path as `ProjectSerializer` and `DatasetSerializer`.
- **`migrate_datasets_to_v2` management command** to backfill existing `Dataset` records into `DatasetV2`.


gsfk (commented):
Looks like a lot of work! First pass is mostly just questions:

- I get a new field `schema_version: "1.0"` Is this the datasets schema version? 
- The current endpoint `/datasets/{id}` returns the discovery config information, but this is missing from `/datasets_v2`. Is this going to be pulled from somewhere else? 
- Is `/datasets_v2` the permanent name for this endpoint, or is it there temporarily to avoid breaking stuff in the pr? We have never really tried versioning our routes, e.g. /datasets/v2/{id}
- similarly I don't understand why `counts` is replaced by `_counts` (`counts` is null for me in the new response for a migrated dataset)
- I had to use `--force-placeholder` to migrate to the new format, and it used the dataset name as a stakeholder name and primary contact name, which may or may not make sense. Not sure if it's sensible to put something here that's more obviously a placeholder rather than an incorrect name (something like "ADD CONTACT NAME").


I also have two puzzles, but they exist already in old datasets format: 

- I'm puzzled by why I have a dataset that shows variants in bento_web but does not have a variants data type in /datasets/{id} or datasets_v2/{id}. I assume the reasoning here is that katsu isn't responsible for variants, but that reasoning makes less sense if katsu is the source of truth for datasets. 
- some routes are duplicated under `/api` for unclear reasons: 
  - `/datasets` and `/api/datasets` both exist
  - `/datasets/{id}` and `/api/datasets/{id}` both exist
  - but `/datasets/{id}/data-types` exists while `/api/datasets/{id}/data-types` does not
 
... `/datasets_v2` has the same duplications. Are some of these routes just forgotten?

chord_metadata_service/chord/serializers.py
```diff
@@ -168,6 +172,66 @@ class Meta:
...
+        if not data.get("identifier"):
+            data = {**data, "identifier": str(uuid.uuid4())}
+        return super().to_internal_value(data)
+
+    def to_representation(self, instance):
+        language = self.context.get("language", "en")
+
+        if language != "en":
+            try:
+                translation = DatasetV2Translation.objects.get(
```
davidlougheed: this is liable to n+1 issues vs accessing it via the instance somehow... is there a clean way to avoid that? maybe via prefetch of translations + access through a manager

chord_metadata_service/metadata/settings.py
```diff
@@ -198,10 +198,16 @@ def str_to_bool(value: str | None) -> bool:
...
+        # **({'django.db.backends': {'level': 'DEBUG'}} if DEBUG else {}),
+        'django.db.backends': {
+            'level': 'WARNING',
+            'propagate': False,
+        },
+        'django.request': _logging_propagate_to_root,
+        'django.channels.server': {'level': 'WARNING'},  # silence in favour of custom access middleware
+        'katsu': {
+            "level": "INFO"
+        }
```
davidlougheed: revert diff

pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.12.0"
-bento-lib = {version = "^15.0.0", extras = ["django"]}
+bento-lib = "^15.1.0"
```
davidlougheed: revert bad change to extras

mock_data/d1.json
```diff
@@ -0,0 +1,196 @@
...
+  "domain": ["Genomics", "Cancer Research"],
+  "program_name": "National Cancer Genomics Program",
+  "extra_properties": {
+    "institution_code": "TU-001",
+    "ethics_approval": "IRB-2022-456",
+    "recruitment_target": 2000,
+    "completion_rate": 0.75,
+    "data_sharing_allowed": true
+  }
+}
```
davidlougheed: missing final newline to be valid unix text file

chord_metadata_service/restapi/urls.py
```diff
@@ -49,6 +52,19 @@
 urlpatterns = [
     path("", include(router.urls)),
     path("", include(batch_router.urls)),
+    # datasets_v2 sub-routes (nested under api/)
```
davidlougheed: why is this manual path list needed here, vs how this was done with old datasets?

chord_metadata_service/chord/management/commands/migrate_datasets_to_v2.py
```diff
@@ -0,0 +1,384 @@
...
+from chord_metadata_service.chord.models import Dataset, DatasetV2
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# DATS → new-schema helpers
+# ---------------------------------------------------------------------------
+
+# Roles available in the new schema that might appear in DATS data
+_KNOWN_ROLES = {
```
davidlougheed: are these from existing dataset JSON we have?

chord_metadata_service/chord/api_views.py
```diff
@@ -256,6 +282,232 @@ async def update(self, request, *args, **kwargs):
         return await sync_to_async(super().update)(request, *args, **kwargs)  # TODO: handle invalid
 
 
+# class DatasetV2ViewSet(CHORDPublicModelViewSet):
```
davidlougheed: rm?

chord_metadata_service/chord/models.py
```diff
@@ -203,6 +205,88 @@ def __str__(self):
         return f"{self.title} (ID: {self.identifier})"
 
 
+# Model, inherted from Mixin
+class DatasetV2(PydanticJSONBModelMixin):
+
+    # --- Mixin configuration ---
+    COLUMN_FIELDS = {
```
davidlougheed, line 212: use double quotes for new code preferably

davidlougheed (commented):
first quick pass

gsfk (commented):
Editing provenance data in bento web creates a new dataset :( ... although can't tell immediately if this isn't bento_web's fault.

chord_metadata_service/restapi/urls.py
SanjeevLakhwani: that was cause I wasn't using @action so didn't automatically register and forgot to do that. But also might be time to consolidate the /api/dataset* and /dataset* into one.

chord_metadata_service/chord/models.py
SanjeevLakhwani: conflict's with Django/PEP guidelines: https://arc.net/l/quote/zizcsqnp

davidlougheed: we don't follow django's style guide.

davidlougheed: PEP does not specify quote prefs.

SanjeevLakhwani: Well I feel like it is too much extra work I would like to follow the django guidelines for now. Unless you really want double quotes on everything.

chord_metadata_service/restapi/urls.py
SanjeevLakhwani: For consolidating I am a bit worried that we might have to change repos outside of public and web. Therefore would leave it for after. Would add a todo for now.

chord_metadata_service/chord/management/commands/migrate_datasets_to_v2.py
SanjeevLakhwani: Was AI gen based on some demo data, I would remove it.

chord_metadata_service/chord/models.py
```diff
@@ -203,6 +206,94 @@ def __str__(self):
...
+        on_delete=models.CASCADE,  # Delete dataset upon project deletion
+        related_name="dv2"
+    )
+
+    identifier = models.CharField(
+        primary_key=True,
+        max_length=128,
+        default=uuid.uuid4,
+        blank=True,
+        help_text="If from PCGL, inherit. Otherwise created in Katsu.",
```
v-rocheleau, line 235: Very happy you thought about that, will make mapping PCGL studies to Bento datasets trivial!

Can you add tests that verify that datasets can be created with an arbitrary `identifier` value?

SanjeevLakhwani: I would add tests later down after the apis are finalized.

chord_metadata_service/chord/migrations/0012_merge_0011_datasetv2_0011_v13_1_0.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 5.2.12 on 2026-03-19 18:00
```
davidlougheed: what's going on with the migrations here?

chord_metadata_service/common/mixins/pydantic_mixin.py
```diff
@@ -0,0 +1,145 @@
...
+
+from django.core.exceptions import ImproperlyConfigured
+from django.db.models.options import Options
+from pydantic import BaseModel, ValidationError as PydanticValidationError
+from rest_framework import serializers
+from rest_framework.settings import api_settings
+from django.db import models
+
+
+class PydanticJSONBModelMixin(models.Model):
```
davidlougheed: this isn't a mixin if it's inheriting from a model and being used as a superclass

chord_metadata_service/common/base_pydantic_jsonb.py
```diff
@@ -0,0 +1,145 @@
...
+    Mixin for models that store some Pydantic fields in a JSONB column.
+
+    Requires:
+        - COLUMN_FIELDS: set of field names stored as actual columns
+        - JSONB_FIELD: name of the JSONField (default: 'other_data')
+        - SCHEMA_CLASS: the Pydantic model class
+    """
+    _meta: Options  # for pylance
+
+    COLUMN_FIELDS: Set[str] = set()
```
davidlougheed: don't use old-style type hinting

chord_metadata_service/common/mixins/pydantic_mixin.py
```diff
@@ -0,0 +1,145 @@
...
+                    setattr(self, f"{k}_id", v)
+                else:
+                    setattr(self, k, v)
+        jsonb_data = {k: v for k, v in data.items() if k not in self.COLUMN_FIELDS}
+        setattr(self, self.JSONB_FIELD, jsonb_data)
+
+
+class PydanticJSONBSerializer(serializers.ModelSerializer):
+    """
+    Generic serializer mixin for models using PydanticJSONBMixin.
```
davidlougheed: wrong class referenced in comment?

chord_metadata_service/chord/models.py
```diff
@@ -203,6 +206,107 @@ def __str__(self):
...
+        return f"{self.dataset_id}: {self.language}"
+
+
+class DatasetV2ScopeAdapter:
+    """Duck-type adapter so DatasetV2 satisfies the ValidatedDiscoveryScope dataset interface."""
+
+    def __init__(self, dataset: "DatasetV2"):
+        self.identifier = dataset.identifier
+        self.project_id = dataset.project_id
+        self.discovery = None  # falls back to project discovery config in ValidatedDiscoveryScope
```
davidlougheed: this whole class seems kinda weird - why can't you store dataset-level discovery config anymore? why can't we just make discovery scope compatible with the new model instead of a duck-typed adapter?

chord_metadata_service/chord/views_search.py
```diff
@@ -379,10 +379,10 @@ async def chord_dataset_search(
 
 @async_api_view(["GET", "POST"])
 @permission_classes([BentoDeferToHandler])
-async def private_dataset_search(request: DrfRequest, dataset_id: str):
+async def private_dataset_search(request: DrfRequest, identifier: str):
```
davidlougheed, line 382: a bit of a pointless diff changing this variable name

chord_metadata_service/discovery/scope.py
```diff
@@ -137,13 +137,13 @@ async def get_discovery_scope(project_id: str | None, dataset_id: str | None) ->
...
         if dataset_id:
-            qs = cm.Dataset.objects.filter(identifier=dataset_id)
+            qs = cm.DatasetV2.objects.filter(identifier=dataset_id)
             if project_id:
                 # check if the dataset exists and belongs to the specified project if project ID is specified;
                 # otherwise, infer the project from the dataset.
                 qs = qs.filter(project_id=project_id)
 
-            dataset = await qs.aget()
+            dataset = cm.DatasetV2ScopeAdapter(await qs.aget())
```
davidlougheed: i don't see the point to this, as mentioned above

chord_metadata_service/experiments/migrations/0017_experiment_dataset_v2.py
```diff
@@ -0,0 +1,69 @@
...
+                        null=True,
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="experiments",
+                        to="chord.datasetv2",
+                    ),
+                ),
+            ],
+            database_operations=[
+                migrations.RunSQL(
+                    sql=_DROP_EXPERIMENT_DATASET_FK
```
davidlougheed, line 62: why did this need to be sql?

chord_metadata_service/resources/cleanup.py
```diff
@@ -21,6 +21,7 @@ async def clean_resources(logger: BoundLogger) -> int:
     resources_referenced = set()
 
     resources_referenced |= await build_id_set_from_model(cm.Dataset, "additional_resources__id")
```
davidlougheed: given that we cant even control the v1 datasets after this update, maybe this line should be removed?

chord_metadata_service/chord/urls.py
```diff
@@ -26,16 +26,42 @@
...
+    path('datasets_v2/<str:identifier>/data-types', views_data_types.dataset_v2_data_type_summary,
+         name="chord-dataset-v2-data-type-summary"),
+    path('datasets_v2/<str:identifier>/data-types/<str:data_type>', views_data_types.dataset_v2_data_type,
+         name="chord-dataset-v2-data-type"),
+    path('datasets_v2/<str:identifier>/translations',
+         DatasetV2TranslationViewSet.as_view({'get': 'list', 'post': 'create'}),
+         name="chord-dataset-v2-translation-list"),
+    path('datasets_v2/<str:identifier>/translations/<str:language>',
+         DatasetV2TranslationViewSet.as_view({'get': 'retrieve', 'put': 'update', 'delete': 'destroy'}),
+         name="chord-dataset-v2-translation-detail"),
```
davidlougheed, line 59: we're starting to get dataset routes mounted two or even three times here - as much as possible i'd stick to single versions of the routes (or at most two – I know we had weirdness before too)

datasetv2 list/create is now available under datasets, datasets_v2, and /api/datasets_v2

chord_metadata_service/chord/api_views.py
```diff
@@ -74,9 +96,24 @@ class ProjectViewSet(CHORDPublicModelViewSet):
...
     serializer_class = ProjectSerializer
 
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context["language"] = _get_preferred_language(self.request)
+        return context
+
+    def retrieve(self, request, *args, **kwargs):
+        response = super().retrieve(request, *args, **kwargs)
+        response["Content-Language"] = _get_preferred_language(request)
```
davidlougheed: if the preferred language isn't available in the model, isn't this then incorrect?

chord_metadata_service/chord/api_views.py
```diff
@@ -256,6 +293,400 @@ async def update(self, request, *args, **kwargs):
...
+
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context["language"] = _get_preferred_language(self.request)
+        return context
+
+    def list(self, request, *args, **kwargs):
+        authz.mark_authz_done(request)
+        response = super().list(request, *args, **kwargs)
+        response["Content-Language"] = _get_preferred_language(request)
```
davidlougheed: idem

chord_metadata_service/chord/urls.py
davidlougheed: especially re: translations, presumably these can be just exposed once under `/api/...`?

davidlougheed (changes requested):

chord_metadata_service/chord/api_views.py
```diff
@@ -256,6 +293,400 @@ async def update(self, request, *args, **kwargs):
...
+            await authz.async_evaluate_one(
+                request,
+                build_resource(project=str(project_id)),
+                P_CREATE_DATASET,
+            )
+        ):
+            return forbidden(request)
+
+        authz.mark_authz_done(request)
+        return await sync_to_async(super().create)(request, *args, **kwargs)
```
v-rocheleau: **request:** would it be possible to make this endpoint behave like a "create or update"?

This is to take into account the fact that we can now create `Datasets` with arbitrary `identifier` values for PCGL `Study` syncing.
It would make ETL a lot simpler to operate if we could always use the same `/datasets` POST endpoint to create or update datasets for PCGL. 
Otherwise we would need to add Katsu API calls and extra logic in ETL jobs to determine the correct endpoint for each Study we try to ingest.

I think this could be implemented with a simple check:
1. Receive a POST request on `/datasets`
2. Use super's `create` if there is no provided `Dataset.identifier` value in the request
3. Use super's `update` if there is a `Dataset.identifier` value in the request **AND** a Dataset with this identifier exists

v-rocheleau (changes requested):
Small request, if possible

chord_metadata_service/chord/api_views.py
v-rocheleau: NVM, this breaks REST.
Will add create-update logic on the ETL side.

chord_metadata_service/chord/urls.py
SanjeevLakhwani: I can change the translations stuff, I am just a bit hesitant, as different services call this and they might break if I dont mount to old or put api. Therefore not the priority of this pr but would be resolved in the next version and would be put under datasets

chord_metadata_service/experiments/migrations/0017_experiment_dataset_v2.py
SanjeevLakhwani: I am going to merge all the migrations, this is just temporary.

chord_metadata_service/chord/migrations/0012_merge_0011_datasetv2_0011_v13_1_0.py
SanjeevLakhwani: would squash migrations soon

chord_metadata_service/chord/management/commands/migrate_datasets_to_v2.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    Build a dict suitable for ProjectScopedDatasetModel.model_validate().
+
+    Raises ValueError if required fields cannot be satisfied and
+    force_placeholder is False.
+    """
+    # --- stakeholders & primary_contact ---
+    stakeholders = _map_creators_to_stakeholders(dataset.creators or [])
+
+    if not stakeholders:
```
davidlougheed: stakeholders can be optional - can this be refactored to not duplicate a fake entry across primary contact and stakeholders?

davidlougheed (changes requested):

v-rocheleau (changes requested):
~~Been testing this quite a bit with ETL and found some edge cases we need to handle.~~

~~Phenopackets/experiments ingestion endpoint still does a `Dataset.identifier` UUID validation, which prevents data ingestion on Datasets that share a PCGL Study ID (non UUID).~~

v-rocheleau:
> ~Been testing this quite a bit with ETL and found some edge cases we need to handle.~
> 
> ~Phenopackets/experiments ingestion endpoint still does a `Dataset.identifier` UUID validation, which prevents data ingestion on Datasets that share a PCGL Study ID (non UUID).~

Nvm my DB was in a weird state I think, just cleaned it and ran the ingestions just fine with ETL and latest pr image pulled

davidlougheed (approved):
lgtm

PR #666, opened 2026-03-09
feat(discovery): return counts by dataset

noctillion:
The discovery endpoint was extended to support per-dataset entity counts, improving granularity for multi-dataset environments. It introduces helpers for fetching and censoring by-dataset counts, and adds them to the response.


chord_metadata_service/discovery/api_views.py
```diff
@@ -558,6 +598,39 @@ async def discovery_endpoint(
...
+    counts_by_dataset_res: dict[str, EntityCountOrBoolResponse] = {}
+
+    ds_level_permissions = await get_data_type_query_permissions(
+        request,
+        list(DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE.values()),
+        dataset_level=True,
+    )
+    has_ds_level_counts_permission = any(p.counts for p in ds_level_permissions.values())
+
+    if has_ds_level_counts_permission:
```
davidlougheed: we will still need to aggregate counts by dataset with project-level permissions, since project-level permissions are *less* permissive than dataset-level permissions. the difference is just that we need to only return summed counts in the case of project-level permissions.

chord_metadata_service/discovery/api_views.py
```diff
@@ -558,6 +598,39 @@ async def discovery_endpoint(
...
+            scope,
+            counts_by_dataset_raw,
+            ds_level_permissions,
+            lg,
+        )
+
+        # When we expose per-dataset counts, recompute the top-level counts from the
+        # censored per-dataset view to avoid residual disclosure (total - sum(others)).
+        if counts_by_dataset_res:
+            count_or_bools_res = aggregate_counts_from_censored_by_dataset(
```
davidlougheed: this needs to be the way we always compute counts everywhere in the code now that we're doing this refactor.

chord_metadata_service/discovery/constants.py
```diff
@@ -16,3 +16,11 @@
     "experiment": ("experiment_result",),
     "experiment_result": (),
 }
+
+ENTITY_TO_DATASET_GROUP_BY = {
+    "phenopacket": "dataset_id__identifier",
+    "individual": "phenopackets__dataset_id__identifier",
+    "biosample": "individual__phenopackets__dataset_id__identifier",
+    "experiment": "biosample__individual__phenopackets__dataset_id__identifier",
```
davidlougheed: biosample doesn't need to go through individual to get to phenopackets

chord_metadata_service/discovery/constants.py
```diff
@@ -16,3 +16,11 @@
     "experiment": ("experiment_result",),
     "experiment_result": (),
 }
+
+ENTITY_TO_DATASET_GROUP_BY = {
+    "phenopacket": "dataset_id__identifier",
```
davidlougheed: you can just do `dataset_id`, dont need `dataset_id__identifier`. same for all below. plus, this info is already in the models via the `get_scope_filters` method, which you can use (albeit maybe we should rename some of the keys in this data structure)

chord_metadata_service/discovery/api_views.py
```diff
@@ -494,6 +500,40 @@ async def discovery_field_response(
...
+    entity_counts_per_entity = await asyncio.gather(
+        *(_get_entity_counts_by_dataset(e) for e in DISCOVERY_ENTITIES)
+    )
+
+    all_datasets: set[str] = set()
+    for ec in entity_counts_per_entity:
+        all_datasets.update(ec.keys())
+
+    res: dict[str, EntityCounts] = {}
+    entities = tuple(DISCOVERY_ENTITIES)
```
davidlougheed: is transforming this into a tuple needed?

davidlougheed (changes requested):
quick first pass

chord_metadata_service/discovery/api_views.py
noctillion: Good catch! Added aggregation per-dataset counts using project-level permissions for protection, without exposing the per-dataset clues in the response.

chord_metadata_service/discovery/api_views.py
noctillion: Agreed- Thanks! field responses are now computed after count_or_bools_res is finalized.

chord_metadata_service/discovery/api_views.py
```diff
@@ -531,11 +570,72 @@ async def discovery_endpoint(
...
         qh = QueryHelper(query, scope, dt_permissions, lg)
         queryset, queried_entities = await qh.get_query_queryset_and_queried_entities(queryset_entity)
-        censored_counts = await qh.get_censored_entity_counts()  # to be used for censoring field responses!
+        # Get both raw counts (for logging) and censored counts; also pre-caches all entity querysets so that
+        # any ValidationError from an invalid query is caught here.
+        counts, count_or_bools_res = await qh.get_censored_entity_counts(return_raw_counts=True)
     except ValidationError as e:
         return await dres.django_validation_error(request, e, lg, "discovery endpoint encountered validation error")
 
+    # -- Counts processing ---------------------------------------------------------------------------------------------
```
davidlougheed: nitpick: rm this header since you have the other header at the same level below

chord_metadata_service/discovery/api_views.py
```diff
@@ -531,11 +570,72 @@ async def discovery_endpoint(
...
+
+    message: str = ""
```
davidlougheed: nitpick: move this definition under the header just below

chord_metadata_service/discovery/api_views.py
```diff
@@ -531,11 +570,72 @@ async def discovery_endpoint(
...
+
+    # -- Per-dataset counts (permissions-dependent) -------------------------------------------------------------------
+
+    counts_by_dataset_res: dict[str, EntityCountOrBoolResponse] = {}
+
+    ds_level_permissions = await get_data_type_query_permissions(
```
davidlougheed: performance: if the scope has a dataset already set, this doesn't need to be fetched since it'll be identical to `dt_permissions`

chord_metadata_service/discovery/api_views.py
```diff
@@ -531,11 +570,72 @@ async def discovery_endpoint(
...
+        request,
+        list(DISCOVERY_ENTITY_NAMES_TO_DATA_TYPE.values()),
+        dataset_level=True,
+    )
+    has_ds_level_counts_permission = any(p.counts for p in ds_level_permissions.values())
+    has_proj_level_counts_permission = any(p.counts for p in dt_permissions.values())
```
davidlougheed: nitpick: this is slightly misnamed + will already be dataset level if a scope dataset is set. you can just leave it unnamed and use the `any(...)` directly below

chord_metadata_service/discovery/api_views.py
```diff
@@ -531,11 +570,72 @@ async def discovery_endpoint(
...
+
+    if has_ds_level_counts_permission or has_proj_level_counts_permission:
+        # Raw per-dataset counts: dataset_id -> {entity -> count}
+        counts_by_dataset_raw: dict[str, EntityCounts] = await discovery_queryset_entity_counts_by_dataset(
+            qh
+        )
+
+        if has_ds_level_counts_permission:
+            # Censor per-dataset counts using dataset-level permissions and expose the breakdown.
```
davidlougheed: this block of code is almost identical to the one below; can you deduplicate it a bit?

chord_metadata_service/discovery/constants.py
```diff
@@ -16,3 +16,11 @@
     "experiment": ("experiment_result",),
     "experiment_result": (),
 }
+
+ENTITY_TO_DATASET_GROUP_BY = {
```
davidlougheed: not needed anymore, right?

chord_metadata_service/discovery/api_views.py
```diff
@@ -494,6 +500,39 @@ async def discovery_field_response(
...
 
+async def discovery_queryset_entity_counts_by_dataset(
+    qqs: QueryHelper,
+) -> dict[str, EntityCounts]:
+    """
+    Returns a dictionary of discovery entity counts grouped by dataset identifier for a given scope/query context.
+    """
+    async def _get_entity_counts_by_dataset(ee: DiscoveryEntity) -> dict[str, int]:
+        qs, _ = await qqs.get_query_queryset_and_queried_entities(ee, validate_field=False)
+        group_by = ENTITY_TO_DATASET_GROUP_BY[ee]
```
davidlougheed: like i mentioned i last review, use the method already defined on the model by using the existing lookup from entity -> model

davidlougheed (changes requested):

davidlougheed:
also, can you merge develop?

chord_metadata_service/discovery/censorship.py
```diff
@@ -121,3 +124,52 @@ async def censor_entity_counts(
...
+
+    return res
+
+
+def aggregate_counts_from_censored_by_dataset(
+    counts_by_dataset: dict[str, EntityCountOrBoolResponse],
+    base_counts: EntityCountOrBoolResponse,
+) -> EntityCountOrBoolResponse:
+
+    aggregated: EntityCountOrBoolResponse = dict(base_counts)
```
davidlougheed: does this need `base_counts` at all?

noctillion: Not really, is very hard to enter this scenario. Good catch. Thanks!

chord_metadata_service/discovery/censorship.py
davidlougheed: all numeric already, don't need the int cast - can just do `sum(values)`

davidlougheed (changes requested):
looking cleaner, one last thing

chord_metadata_service/discovery/censorship.py
noctillion: perfect!

noctillion:
> looking cleaner, one last thing

Thanks for the review!

davidlougheed (approved):
lgtm. when i get the chance i'll make a ticket for implementing this type of censorship everywhere in katsu.

noctillion:
> lgtm. when i get the chance i'll make a ticket for implementing this type of censorship everywhere in katsu.

Awesome! Sounds great! Thanks!

PR #669, opened 2026-03-13
fix(discovery): properly check FTS discovery query perms

davidlougheed:
* fts queries should no longer work without `query:data` permissions at the API level


gsfk (approved):
Looks good. I guess I missed this the first time?

PR #686, opened 2026-04-13
feat(experiments): add storage_uri and storage_server fields

davidlougheed:
These fields are for annotating a file's storage location on a server (not necessarily an accessible location, but more for use by ICHANGE/similar to find their own files.)


noctillion (approved):
LGTM! It works well

PR #688, opened 2026-04-15
fix: normalize vital_status fields in get_or_create query

noctillion:
fix MultipleObjectsReturned when ingesting phenopackets with partial vital_status fields

### DIFF
--- a/chord_metadata_service/chord/ingest/phenopackets.py
+++ b/chord_metadata_service/chord/ingest/phenopackets.py
@@ -111,7 +111,10 @@ def update_or_create_subject(subject: dict) -> pm.Individual:
 
     vital_status: VitalStatus | None = None
     if vital_status_data := subject.get("vital_status"):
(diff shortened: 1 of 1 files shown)


chord_metadata_service/chord/ingest/phenopackets.py
```diff
@@ -111,7 +111,10 @@ def update_or_create_subject(subject: dict) -> pm.Individual:
 
     vital_status: VitalStatus | None = None
     if vital_status_data := subject.get("vital_status"):
-        vital_status, _ = VitalStatus.objects.get_or_create(**vital_status_data)
+        vs_query = {"status": vital_status_data.get("status", "")}
```
davidlougheed: "" (empty string) is not a valid vital status, default to `UNKNOWN_STATUS`

davidlougheed (changes requested):

davidlougheed (approved):
thanks for catching this issue! lgtm

PR #690, opened 2026-04-24
fix spelling of scChIC-seq in experiment type

gsfk:
Spelling fix for scChIC-seq, see, eg, [here](https://pmc.ncbi.nlm.nih.gov/articles/PMC7187538/) for a reference.


gsfk:
closing, should be "scHi-C" ie single-cell Hi-C, will fix in a separate pr

PR #691, opened 2026-04-24
spelling fix for single cell Hi-C

gsfk:
single cell Hi-C should be "scHi-C", see e.g. [here](https://advanced.onlinelibrary.wiley.com/doi/10.1002/advs.202412232) for a reference.


noctillion (approved):
LGTM!

PR #705, opened 2026-06-08
feat(chord): validate translation constraints on create/update

SanjeevLakhwani:
## Summary
Redmine ticket [2830](https://redmine.c3g-app.sd4h.ca/issues/2830)

- Reject translations where `primary_contact` or any `stakeholders` entry has different `roles` than the canonical (English) dataset
- Reject translations where any list field (`keywords`, `stakeholders`, `taxa`, `links`, `publications`, `logos`, `counts`, `resources`, `participant_criteria`, `domain`, `funding_sources`) grows beyond the canonical size
- Validation runs in `DatasetTranslationSerializer.to_internal_value()` after Pydantic validation, covering both create and update paths; looks up the canonical dataset by `translation.identifier` so it works regardless of how the serializer is instantiated
- 11 new tests added covering rejection and acceptance cases for both rules on both POST and PUT


chord_metadata_service/chord/serializers.py
```diff
@@ -126,6 +127,84 @@ def update(self, instance, validated_data):
...
+                errors[field] = [
+                    f"Translation must have the same number of items in '{field}' as canonical "
+                    f"(canonical has {len(c_val)}, translation has {t_len})."
+                ]
+
+    # Rule 3: if an immutable field is present in the translation it must match canonical exactly
+    for field in _IMMUTABLE_FIELDS:
+        c_val = getattr(canonical, field, None)
+        t_val = getattr(translation, field, None)
+        if t_val is not None and t_val != c_val:
```
davidlougheed, line 204: `discovery` is an immutable field, but is an object and so cannot be compared. however this is handled elsewhere anyway since `discovery `cannot be in a translation - should the discovery check be here in a separate rule instead? forbidding translations from ever having this?

chord_metadata_service/chord/serializers.py
```diff
@@ -126,6 +127,84 @@ def update(self, instance, validated_data):
...
+def _roles_for(contact) -> list:
+    """Return roles list for a contact, defaulting to empty list."""
+    return getattr(contact, "roles", []) or []
+
+
+_IMMUTABLE_FIELDS = frozenset({"version", "release_date", "last_modified", "study_status", "study_context",
+                               "discovery", "pcgl_dac_id"})
+
+
+def _check_translation_constraints(translation: ProjectScopedDatasetModel):
```
davidlougheed, line 141: this is done in serialization, but I feel like it should be done in the actual model save method instead to prevent invalid translations from *ever* being created, not just in the API (e.g., in a hypothetical CLI.) this is my general preference for MVC apps, to have validation closer to the data. however, if there's a reason this is not as good, i'm fine with being convinced otherwise.

chord_metadata_service/chord/serializers.py
```diff
@@ -126,6 +127,84 @@ def update(self, instance, validated_data):
         return instance
 
 
+def _roles_for(contact) -> list:
+    """Return roles list for a contact, defaulting to empty list."""
+    return getattr(contact, "roles", []) or []
+
+
+_IMMUTABLE_FIELDS = frozenset({"version", "release_date", "last_modified", "study_status", "study_context",
```
davidlougheed, line 137: see my comment below - should discovery be here?

chord_metadata_service/chord/serializers.py
SanjeevLakhwani: The model is just taking a JSON field, I don't see how close to the model we can get with that. Given that in the past we discussed about the effort to be done in implementing this and was asked to be low. It is known that a tech debt is being added. My idea for the future for this is to separate main model from translation. Everything gets serialized into a different translation model, which allows to not store duplicates of data that shouldn't be and have validation against canonical.

PR #706, opened 2026-06-11
make minimum and maximum optional for discovery autobins

gsfk:
Remove the requirements for `minimum` and `maximum` for automatically generated bins in discovery config. Requires [bento_lib pr 343](https://github.com/bento-platform/bento_lib/pull/343)

The behaviour is unchanged:

- values in the data that are less than minimum or greater than maximum are ignored (not counted in charts) 
- the leftmost bin is everything less than taper_left unless minimum = taper_left
- the rightmost bin is everything greater than taper_right unless maximum = taper right

This matches the behaviour of so-called "manual" bins (where we consider the values for the left-most and right-most bins instead of "taper" values). It also allows for minimum and maximum to be used as meaningful limits in the search UI (rather than, e.g. having awkward hacks such as max age = 125 just so the bins will work)


SanjeevLakhwani (approved):
lgtm

davidlougheed (approved):
lgtm

PR #712, opened 2026-07-06
refactor(restapi): field registry for CSV export renderers

SanjeevLakhwani:
## Summary
- Replaces the hand-written `get_columns()`/`get_dicts()` in each `KatsuCSVRenderer` subclass (Individual, Phenopacket, Biosample, Experiment, ExperimentResult) with a declarative `field_registry: dict[str, FieldSpec]` per entity.
- `KatsuCSVRenderer` base class now derives columns and row-dicts generically from that registry — subclasses just declare `field_registry` + `file_name` + `get_model_serializer()`.
- Output is unchanged: same columns, same order, same values, same headers, same error/pagination handling.

## Future goal (not in this PR)
- Expose registry keys via a small `export_fields` action per viewset so the Bento UI can discover selectable columns for a picker.
- Accept a selected-fields list on export requests (query param or POST body) instead of always returning every registered column.
- Optionally: project/dataset-level saved export templates (named field-set presets per org).
- XLSX export with one sheet per entity for nested data (biosamples/experiments per individual) instead of the current ad hoc `"; "`-joined flattening.
- Decouple full-result export from pagination for large datasets (today, direct viewset CSV export only covers the current page; only `discovery_matches` with `_page_size=0` returns everything).

### DIFF
--- a/chord_metadata_service/restapi/api_renderers.py
+++ b/chord_metadata_service/restapi/api_renderers.py
@@ -128,11 +128,17 @@ class FieldSpec:
 
 
 def get_path(row: dict, *path: str, default: Any = None) -> Any:
-    """Walk a chain of dict keys, short-circuiting to `default` if any step is missing or not a dict."""
+    """
+    Walk a chain of dict keys, short-circuiting to `default` if a step is missing/None. Raises TypeError if a step
+    resolves to a non-None, non-dict value while path segments remain - that means the field registry's path
+    doesn't match the actual row shape, which is a bug, not a normal absent-value case.
+    """
     val: Any = row
     for key in path:
-        if not isinstance(val, dict):
+        if val is None:
             return default
+        if not isinstance(val, dict):
+            raise TypeError(f"get_path: expected dict while resolving {key!r}, got {type(val).__name__}")
         val = val.get(key)
     return default if val is None else val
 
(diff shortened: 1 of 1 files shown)


chord_metadata_service/restapi/api_renderers.py
```diff
@@ -119,24 +120,49 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+    """Declares how to pull one exportable column's value out of a serialized row (a dict, from a DRF serializer)."""
+
+    getter: Callable[[dict], Any]
+
+
+def get_path(row: dict, *path: str, default: Any = None) -> Any:
+    """Walk a chain of dict keys, short-circuiting to `default` if any step is missing or not a dict."""
+    val: Any = row
+    for key in path:
+        if not isinstance(val, dict):
```
davidlougheed: q: is there any case where we should raise instead of suppressing the error here? for None, i get just returning a default, but perhaps in other cases we should raise?

davidlougheed (commented):
overall i like this direction, although in the future we should really figure out a way to not be duplicating "lists of fields of entities" so many times everywhere in katsu (schemas, serializers, models, pydantic models, etc.)

chord_metadata_service/restapi/api_renderers.py
SanjeevLakhwani: Good catch; fixed in 15c6e3b6. Now: value is `None`/missing → still returns `default` (normal case). Value present but non-dict while path segments remain → raises `TypeError`, since that means the field spec's path doesn't match the actual row shape (a bug in the registry, not an absent value).

davidlougheed (approved):
looks good, thank you

PR #714, opened 2026-07-09
feat(restapi): export_fields + field selection for CSV export, incl. discovery_matches

SanjeevLakhwani:
## Summary
- Adds a GET `export_fields` action per CSV-enabled viewset (`individuals`, `batch/individuals`, `batch/biosamples`, `batch/experiments`) returning `{key, label}` pairs from the field registry, so the Bento UI can discover selectable columns for a picker.
- CSV export requests now accept a `fields` list (query param on GET, body field on POST) to restrict which columns are returned, instead of always returning every registered column. Unrecognized keys 400 with the standard bento_lib error body rather than being silently dropped.
- Extends the same discovery/selection support to `discovery_matches`, the endpoint the Bento UI actually uses for CSV export:
  - New `discovery_matches_export_fields?_entity=...` endpoint, mirroring the viewset `export_fields` actions.
  - `discovery_matches` accepts an underscore-prefixed `_fields` selector (plain `fields` would be swallowed as a bogus discovery filter field, since this endpoint treats any non-underscore query param as one) with the same unknown-key 400 behavior.
- Adds an XLSX export option to `discovery_matches` (`_format=xlsx`): a flat single-sheet workbook with the same columns/values/ordering as the CSV export, same `_fields` selector and validation, same `discovery_matches_export_fields` picker (CSV and XLSX share columns, so no separate picker endpoint was needed). Adds `openpyxl` as a dependency.
- Fixes two latent bugs this surfaced:
  - `KatsuCSVRenderer`'s non-200 branch built a fresh `HttpResponse` whose headers DRF discards when rendering happens through its own `Response` pipeline, so error bodies were served with a stale `text/csv` Content-Type instead of JSON.
  - The same branch unconditionally indexed `renderer_context["response"]`, which crashes when the renderer is invoked directly with only a `{"request": ...}` context (as `discovery_matches` now does to make `_fields` available).
- Output for the "return everything" case is unchanged across all endpoints.

Builds on #712 (field registry refactor).

## Next steps (not in this PR)
- Optionally: project/dataset-level saved export templates (named field-set presets per org).
- XLSX for the 4 direct viewsets, and/or multi-sheet-per-entity XLSX for nested data (biosamples/experiments per individual) instead of the current ad hoc `"; "`-joined flattening - only discovery_matches got XLSX in this PR, matching where it's actually used.
- Decouple full-result export from pagination for `/individuals` (GET) and `batch/individuals` (POST) - both still cap out at `page_size=10000` via their pagination classes. (`batch/biosamples` and `batch/experiments` already export the full filtered queryset via their POST `create()` path, which bypasses pagination entirely; only the GET `list()` path on those two stays paginated, and `discovery_matches` already supports full export via `_page_size=0`.)


chord_metadata_service/restapi/api_renderers.py
```diff
@@ -186,23 +261,74 @@ def _generate_csv_response(self, data: list[dict[str, str]]):
...
+
+    file_name: str = "data.xlsx"
+
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        columns = self.get_columns(renderer_context)
+        rows = self.get_dicts(data, columns) if data else []
+
+        wb = Workbook()
+        ws = wb.active
+        ws.title = "Export"
```
davidlougheed: I think we could make room for future support of multi-sheet exports by naming the sheet after the entity being exported instead if that makes sense?

gsfk:
Still testing, but a few comments: 
- ~~pagination already seems decoupled from csv export, contrary to the comment in "Next Steps" above~~ (my bad, I was not testing results > 10K, I would expect issues elsewhere with that much data)
- there are a few errors / wonky issues in the existing CSV export that are carried over here, could be fixed here or separately.

chord_metadata_service/restapi/api_renderers.py
SanjeevLakhwani: I agree, implemented

gsfk (approved):
Working well for me. Some odd content in a few export fields but that's arguably out of scope here.

chord_metadata_service/experiments/api_views.py
```diff
@@ -99,27 +99,46 @@ async def _get_filtered_queryset(self, ids_list: list[str] | None = None):
 
         return queryset.select_related(*EXPERIMENT_SELECT_REL).prefetch_related(*EXPERIMENT_PREFETCH).order_by("id")
 
+    @async_to_sync
+    async def _get_filtered_queryset(self, ids_list: list[str] | None = None):
```
davidlougheed, line 103: what is the purpose of this, given it just wraps `_filtered_queryset`?

SanjeevLakhwani: Yeah but a sync wrapper?

davidlougheed (approved):
lgtm

PR #716, opened 2026-07-13
fix(restapi): enforce CURIE pattern on ontology class id at ingest

SanjeevLakhwani:
## Summary
- Non-CURIE ontology IDs (e.g. raw OBO IRIs like `http://purl.obolibrary.org/obo/OBI_0002117`) passed ingest validation but failed later at search/discovery time, since `bento_lib.ontologies.models.OntologyClass` enforces a strict CURIE pattern that katsu's own ingest schema never checked.
- `ONTOLOGY_CLASS["id"]` in `restapi/schemas.py` now reuses `bento_lib.ontologies.models.CURIE_PATTERN` as a single source of truth, so ingest and search validation agree.
- Fixed test fixtures that were themselves relying on the validation gap (underscore-style OBO ids, a raw EFO IRI).

### DIFF
--- a/chord_metadata_service/chord/tests/example_experiment.json
+++ b/chord_metadata_service/chord/tests/example_experiment.json
@@ -6,7 +6,7 @@
       "study_type": "Epigenomics",
       "experiment_type": "Other",
       "experiment_ontology": {
-          "id": "http://www.ebi.ac.uk/efo/EFO_0002692",
+          "id": "EFO:0002692",
           "label": "ChIP-seq"
       },
       "library_strategy": "ChIP-Seq",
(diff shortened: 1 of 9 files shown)


gsfk (commented):
works as intended, although two comments:

- why is there a duplicate ONTOLOGY_CLASS rather than using bento_lib [OntologyClass](https://github.com/bento-platform/bento_lib/blob/922adecea05b5d0594111be828c3c2bf6f2fb668/bento_lib/ontologies/models.py#L96-L109)? Maybe this is a historical issue. 
- needs changes to broken ontologies in `examples/experiments.json` ~~and presumably also [here](https://github.com/bento-platform/katsu/blob/f991ddaffafdf560ebfbb926bd453dc51908c8b0/chord_metadata_service/phenopackets/tests/test_api.py#L37)~~.

gsfk (approved):

PR #717, opened 2026-07-16
feat(discovery): emit machine-readable date bin labels

SanjeevLakhwani:
- Date field bins in discovery/discovery_matches responses now carry a machine-readable `"yyyy-mm"` label instead of a server-formatted `"Mon YYYY"` string (e.g. `"2021-01"` instead of `"Jan 2021"`).
- Front-end already receives the binning unit via existing `definition.config.bin_by` ("month") and can now format the display label itself instead of relying on server-side formatting.
- Date bin filter parsing updated to accept `"yyyy-mm"` directly (old `"Mon YYYY"` filter format no longer accepted).
- Numeric and categorical field bins are unchanged.


chord_metadata_service/discovery/fields_utils.py
```diff
@@ -157,15 +156,11 @@ def _year_month_from_month_offset(start_year: int, month_offset: int) -> tuple[i
     return start_year + month_offset // 12, (month_offset + 1) % 12 or 12
 
 
-def _month_label(year: int, month: int) -> str:
```
davidlougheed, line 160: feels good

davidlougheed (approved):
lgtm, merge once public is ready

PR #718, opened 2026-07-20
feat(experiments): add batch/experimentresults endpoint (CSV + XLSX export)

SanjeevLakhwani:
## Summary
- `experiment_result` was the only entity with CSV/XLSX export (via `discovery_matches`) but no batch-by-ID endpoint of its own, unlike `individuals`/`biosamples`/`experiments`.
- Adds `ExperimentResultBatchViewSet` at `/api/batch/experimentresults`, mirroring `ExperimentBatchViewSet`:
  - `GET` - list all (unfiltered)
  - `POST {"id": [...]}` - filter to a specific ID list (unknown IDs silently dropped, matching sibling batch endpoints)
  - `POST {"format": "csv"|"xlsx", "fields": [...]}` - CSV or XLSX export with column selection
  - `GET /export_fields` - `{key, label}` column picker, same `csv_fields_error_response` 400-on-unknown-field validation as the other batch endpoints
- Also adds XLSX as an export option on this endpoint (`format=xlsx`, same `fields` selector/validation as CSV) - previously XLSX only existed on `discovery_matches`.

## Bug fixes surfaced along the way
Wiring `KatsuXLSXRenderer` into a real viewset (rather than only being called directly from `discovery_matches`) exposed two bugs in it:
- `render()` never unwrapped paginated `{"count", "results": [...]}` responses before iterating rows, unlike `KatsuCSVRenderer` - crashed on any paginated batch endpoint response.
- `BaseRenderer`'s default `charset='utf-8'` made DRF append `"; charset=utf-8"` to the XLSX Content-Type header once content-type assembly went through the normal `Response.rendered_content` path instead of being invoked directly.

Fixed by refactoring the shared `render()` flow (column selection, paginated-dict unwrapping, non-200 JSON error fallback) into the `FieldRegistryRenderer` mixin, with format-specific classes only implementing `_build_response(rows, columns)`. Same charset fix applied to `PassThruXLSXRenderer` for consistency.

## Notes
- `ExperimentResult` uses an integer PK (unlike the string IDs on `Individual`/`Biosample`/`Experiment`), so passing a non-numeric value in `id` raises rather than being silently ignored - pre-existing model behavior, not something this endpoint changes. Covered in tests with an out-of-range int instead.
- Based on `feature/csv-export-fields-picker` (not `develop`) since it reuses that branch's `ExperimentResultCSVRenderer`/`ExperimentResultXLSXRenderer`/`field_choices`/`csv_fields_error_response` infra.

### DIFF
--- a/chord_metadata_service/experiments/api_views.py
+++ b/chord_metadata_service/experiments/api_views.py
@@ -16,6 +16,7 @@ from chord_metadata_service.restapi.api_renderers import (
     PhenopacketsRenderer,
     ExperimentCSVRenderer,
     ExperimentResultCSVRenderer,
+    ExperimentResultXLSXRenderer,
     csv_fields_error_response,
 )
 from chord_metadata_service.restapi.constants import MODEL_ID_PATTERN
@@ -178,7 +179,12 @@ class ExperimentResultBatchViewSet(BentoAuthzScopedModelGenericListViewSet):
 
     serializer_class = ExperimentResultSerializer
     pagination_class = BatchResultsSetPagination
-    renderer_classes = (*api_settings.DEFAULT_RENDERER_CLASSES, PhenopacketsRenderer, ExperimentResultCSVRenderer)
+    renderer_classes = (
+        *api_settings.DEFAULT_RENDERER_CLASSES,
+        PhenopacketsRenderer,
+        ExperimentResultCSVRenderer,
+        ExperimentResultXLSXRenderer,
+    )
     content_negotiation_class = FormatInPostContentNegotiation
 
     data_type = DATA_TYPE_EXPERIMENT
--- a/chord_metadata_service/experiments/api_views.py
+++ b/chord_metadata_service/experiments/api_views.py
@@ -15,6 +15,7 @@ from chord_metadata_service.discovery.scope import get_request_discovery_scope
 from chord_metadata_service.restapi.api_renderers import (
     PhenopacketsRenderer,
     ExperimentCSVRenderer,
+    ExperimentResultCSVRenderer,
     csv_fields_error_response,
 )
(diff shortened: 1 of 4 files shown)


gsfk (changes requested):
Trouble with `/batch/individuals` not recognizing the fields emitted from `/discovery_matches_export_fields`. Attempts at customizing fields return 400 Bad Request with a message, eg:

```
"errors": [
        {
            "message": "unknown export field(s): subject_id, subject_sex, subject_taxonomy, biosamples, created_by, submitted_by, dataset"
        }

```

SanjeevLakhwani:
> Trouble with `/batch/individuals` not recognizing the fields emitted from `/discovery_matches_export_fields`. Attempts at customizing fields return 400 Bad Request with a message, eg:
> 
> ```
> "errors": [
>         {
>             "message": "unknown export field(s): subject_id, subject_sex, subject_taxonomy, biosamples, created_by, submitted_by, dataset"
>         }
> ```

`/discovery_matches_export_fields` defaults to phenopacket fields, which don't match `/batch/individuals` own field set. Use `GET /batch/individuals/export_fields` to get the correct field list for that endpoint.

gsfk:
~~pretty sure I produced this error in bento_public, although I don't see it now, so maybe not.~~

yeah, this is a bento_public issue, I'll note something in the other pr

gsfk (approved):

davidlougheed (approved):
lgtm

gsfk (changes requested):
`/batch/individuals` returning json instead of xlsx for xlsx requests

hmm... I guess this goes with the other pr, already merged

gsfk (approved):

PR #734, opened 2026-08-18
fix(geo): make GeoLocation properties optional

davidlougheed:
aligns better with GeoJSON/GeoLocation spec


gsfk:
Ingestion works fine, note that when when serializing, katsu will insert 

`properties: {}`

for GeoJSONs where `properties` is absent. Is this okay?

davidlougheed:
yeah that's fine, just wanted to be able to ingest geojson points without specified properties

gsfk (approved):

